PR #11, opened 2019-03-28
Add the deleteMultiSig operation to the kmd client.

adjicf:


JasonWeathersby:
Any chance you could resubmit the pr? we have a cla in the process now

PR #12, opened 2019-03-29
Eric/b64 update

egieseke:
Added conversion of noteb64 to note byte array for SDK methods block, transaction By Address, and pending Transactions.  Please do some additional testing.


src/client/algod.js
```diff
@@ -74,6 +91,11 @@ function Algod(token, baseServer = "http://r2.algorand.network", port = 4180) {
     this.transactionByAddress = async function (addr, first, last) {
         if (!Number.isInteger(first) || !Number.isInteger(last)) throw Error("first and last rounds should be integers");
         let res = await c.get("/v1/account/" + addr + "/transactions", {'firstRound': first, 'lastRound': last});
+        if (res.statusCode === 200) {
+          for(var i = 0; i < res.body.transactions.length; i++) {
+            res.body.transactions[i] = noteb64ToNote(res.body.transactions[i], "base64");
```
rotemh: `noteb64ToNote` takes only 1 argument

src/client/algod.js
```diff
@@ -34,6 +34,18 @@ function Algod(token, baseServer = "http://r2.algorand.network", port = 4180) {
         return res.body;
     };
 
+    /**
+     * Takes an object and convert its note field to Buffer, if exist.
+     * @param o
+     * @returns {*}
+     */
+    function noteb64ToNote(o) {
```
rotemh: Can we move this function to the top of the file?

src/client/algod.js
egieseke: Fixed parameter count.

src/client/algod.js
egieseke: Moved to the top of the file.

PR #13, opened 2019-04-11
Ensure thrown errors from algod.js have error.message defined.

EvanJRichard:
## Summary
Problem statement from @JasonWeathersby :
> When using the JS SDK the returned error message is not set:
> 
> let tx = (await algodclient.transactionInformation( "NJY27OQ2ZXK6OWBN44LE4K43TA2AV3DPILPYTHAJAMKIVZDWTEJKZJKO4A", "EXQVLGRRARXG5KG5EBZVVJD3GK2GI5S4PET33OKN36S42KD65ZRQ" ));
> 
> Wrap with try catch and print out err.message. It will be undefined. err prints though.

This PR proposes to wrap algod.js calls in a try/catch. When catching an error, a new error is re-thrown using the extracted error message, so that `err.message` is defined, as expected.

### Testing
Used minimal example provided by Jason to test if `err.message` was defined. Also hand-tested other functions such as `sendRawTransaction`, which needed different re-throw logic.

### DIFF
--- a/src/client/algod.js
+++ b/src/client/algod.js
@@ -21,8 +21,12 @@ function Algod(token, baseServer = "http://r2.algorand.network", port = 4180) {
      * @returns {Promise<*>}
      */
     this.status = async function () {
-        let res = await c.get("/v1/status");
-        return res.body;
+        try {
+            let res = await c.get("/v1/status");
+            return res.body;
+        } catch (e) {
+            throw Error(e.response.error.message)
+        }
     };
 
     /**
@@ -30,8 +34,12 @@ function Algod(token, baseServer = "http://r2.algorand.network", port = 4180) {
      * @returns {Promise<*>}
      */
     this.healthCheck = async function () {
-        let res = await c.get("/health");
-        return res.body;
+        try {
+            let res = await c.get("/health");
+            return res.body;
+        } catch (e) {
+            throw Error(e.response.error.message)
+        }
     };
 
     /**
@@ -42,8 +50,12 @@ function Algod(token, baseServer = "http://r2.algorand.network", port = 4180) {
      */
     this.statusAfterBlock = async function (roundNumber) {
         if (!Number.isInteger(roundNumber)) throw Error("roundNumber should be an integer");
-        let res = await c.get("/v1/status/wait-for-block-after/" + roundNumber);
-        return res.body;
+        try {
+            let res = await c.get("/v1/status/wait-for-block-after/" + roundNumber);
+            return res.body;
+        } catch (e) {
+            throw Error(e.response.error.message)
+        }
     };
 
     /**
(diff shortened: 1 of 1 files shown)


src/client/algod.js
```diff
@@ -21,17 +21,25 @@ function Algod(token, baseServer = "http://r2.algorand.network", port = 4180) {
      * @returns {Promise<*>}
      */
     this.status = async function () {
-        let res = await c.get("/v1/status");
-        return res.body;
+        try {
```
rotemh: Seems like it worth trying to put the try catch mechanism in a separate function.

src/client/algod.js
```diff
@@ -21,17 +21,25 @@ function Algod(token, baseServer = "http://r2.algorand.network", port = 4180) {
...
      * @returns {Promise<*>}
      */
     this.healthCheck = async function () {
-        let res = await c.get("/health");
-        return res.body;
+        try {
+            let res = await c.get("/health");
+            return res.body;
+        } catch (e) {
+            throw Error(e.response.error.message)
```
rotemh: You should throw back the entire error. You lose a lot of information by wrapping just the message

EvanJRichard: @rotemh It turned out that the error itself was being propagated up just fine, it was not being swallowed; the ticket's problem was that the thrown error was not of the expected format. (I.e. in the example, if one catches the resulting error `e`, one can `console.log(e)`, but `console.log(e.message)` will be undefined.)

I'll look into a version that defines `e.message` while preserving the rest of the information. :+1:

src/client/algod.js
EvanJRichard: @rotemh I initially agreed but after trying this, I think it might be cleaner/shorter as-is. For context, I was trying to implement some sort of 
```
tryAndCatchWrapper = function(func) {
    try {
        func()
    } catch (e) {
      //context-dependent re-throw logic
    }
}
```
And then each `algod.js` function would pass its logic as `func`, but that quickly ended up more verbose and confusing than just using `try`/`catch`.

src/client/algod.js
```diff
@@ -21,17 +21,27 @@ function Algod(token, baseServer = "http://r2.algorand.network", port = 4180) {
      * @returns {Promise<*>}
      */
     this.status = async function () {
-        let res = await c.get("/v1/status");
-        return res.body;
+        try {
+            let res = await c.get("/v1/status");
+            return res.body;
+        } catch (e) {
+            e.message = e.response.error.message;
```
rotemh: Consider simply returning the entire error obj `throw e.response.error`. Right now the code breaks abstractions and makes some pieces of the error hidden (such as the error code)

src/client/algod.js
rotemh: I thought of wrapping the entire get functionality so you'd pass just the path, but I accept that. It's not a big deal.

src/client/algod.js
EvanJRichard: The proposed change adds an extra field to the `e` object with no deletion, whereas throwing just `e.response.error` would remove much (>90%, as measured by field count) of the error data. 

Here's an example error object that comes out from the currently-proposed change: https://gist.github.com/EvanJRichard/e3872fb953dbdb68607a18e2e0e0d385. It's the same error object that was being returned, with all data preserved (including error code), with an additional field `message` added.

rotemh: Now, when I understand the purpose of this PR, I think it should be closed. 
The information exist at the top level of the response returned (by examine your example and some that I ran personally). 

The `message` that the PR surfaces up already exist just under the name `text` and the statusCode is under the name `Status` (Although wasn't surfaced)

EvanJRichard:
Closing PR as `wontfix` (see discussion).

PR #21, opened 2019-05-14
Add multisig support.

Vervious:
Add support for the following functionality:
- `let partiallySignedTxBlob = algosdk.signMultisigTransaction(tx, multisigparams, secretKey)` to begin a new multisig txn
- `let partiallySignedTxBlob = algosdk.appendSignMultisigTransaction(multisigBlob, multisigparams, secretKey)` to append signature
- `let partiallySignedTxBlob = algosdk.mergeMultisigTransactions([multisigBlob1, multisigBlob2, ...])` to merge multisigs

Each function takes in multisig metadata, e.g.
```
const multisigparams = {
    version: 1,
    threshold: 2,
    addrs: [
        "DN7MBMCL5JQ3PFUQS7TMX5AH4EEKOBJVDUF4TCV6WERATKFLQF4MQUPZTA",
        "BFRTECKTOOE7A5LHCF3TTEOH2A7BW46IYT2SX5VP6ANKEXHZYJY77SJTVM",
        "47YPQTIGQEO7T4Y4RWDYWEKV6RTR2UNBQXBABEEGM72ESWDQNCQ52OPASU",
    ],
};
```
where appropriate. This substitutes for the corresponding functionality of needing to create a multisig account when using `goal`; instead, the user passes in these parameters every time.

Also enable generating multisig public addresses from preimage parameters:
- `let multisigAddr = algosdk.multisigAddress(multisigparams)`


src/multisig.js
```diff
@@ -0,0 +1,184 @@
+const nacl = require('./nacl/naclWrappers');
+const address = require('./encoding/address');
+const encoding = require('./encoding/encoding');
+const txnBuilder = require('./transaction');
+
+/**
+ Utilities for manipulating multisig transaction blobs.
+ */
+
+const ERROR_MULTISIG_MERGE_LESSTHANTWO = new Error("Not enough multisig transactions to merge. Need at least two");
```
rotemh: This is an arbitrary constraint. Although makes sense, I think that threshold of 1 is perfectly reasonable.

src/multisig.js
```diff
@@ -0,0 +1,184 @@
...
+    return new Uint8Array(encoding.encode(sTxn));
+}
+
+/**
+ * mergeMultisigTransactions takes a list of multisig transaction blobs, and merges them.
+ * @param multisigTxnBlobs a list of blobs representing encoded multisig txns
+ * @returns typed array msg-pack encoded multisig txn
+ */
+function mergeMultisigTransactions(multisigTxnBlobs) {
+    if (multisigTxnBlobs.length < 2) {
```
rotemh: same as above. Also, seems like you don't enforce the creation of threshold > 1

src/multisig.js
Vervious: Oh, this is something else entirely - we need at least two multisig transactions in order to merge them into one.

src/multisig.js
Vervious: I think its fairly reasonable two require two transactions in order to merge them.

rotemh: if the structure allows you to send one by itself, it's fine.

PR #23, opened 2019-05-23
Added ability to pass your own randomization function

twtaylorbitgo:
Account generation doesn't allow your own custom function for randomization


src/main.js
```diff
@@ -208,6 +208,7 @@ module.exports = {
     signBid,
     encodeObj,
     decodeObj,
+    nacl,
```
twtaylorbitgo: this actually was my original approach and will be removed

rotemh:
As discussed, we rather avoid this functionality at this point.

PR #26, opened 2019-05-23
Add logic to encode secret key to plaintext seed

CapnMigraine:
Add ability to convert a secret key to a base32 encoded seed.


rotemh:
Hi, 

Thanks for the PR. Unfortunately, we think that this feature cannot be merged to our SDK for two main reasons - The first, we would like to not turn the SDK into everything-but-the-kitchen-sink solution. The second, we think b32 will be confusing since we use it for addresses and don't want people to mix the two.

PR #35, opened 2019-07-22
signMultisigTransaction and exportMultisig fix

mjiang102628:
- removed wallet password from exportMultisig
- changed delete to post in signMultisigTransaction
- changed "MultiSig" occurences to "Multisig"


Vervious:
We should have probably bumped the version due to the renaming

PR #47, opened 2019-08-22
add signBytes, transactionsByAddressAndDate, pendingTransaction Information

mjiang102628:
- add signing and verifying bytes
- add functions to convert from microalgos to algos and back
- add function to get transactions by address and date
- add function to get pending transaction information
- add maxTxns parameter for getting transactions by address

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -7,6 +7,7 @@ const multisig = require('./multisig');
 const bidBuilder = require('./bid');
 const algod = require('./client/algod');
 const kmd = require('./client/kmd');
+const utils = require("./utils/utils");
 
 let Algod = algod.Algod;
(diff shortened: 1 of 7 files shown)


src/main.js
```diff
@@ -199,13 +228,34 @@ function decodeObj(o) {
...
+    return Big(microalgos).div(1e6)
+}
+
+/**
+ * algosToMicroalgos converts algos to microalgos
+ * @param algos number, string, or Big
+ * @returns Big
+ */
+function algosToMicroalgos(algos) {
+    return Big(algos).times(1e6).round()
```
rotemh: Consider using `isSafeInteger` function and avoid importing `Big`.

PR #49, opened 2019-09-11
Added additional header capability to algod client and it's methods

fxgamundi:
Adds support for multiple different headers on the algod client and in all of the internal methods.
Backwards compatible for existing code, both normal Algod string tokens and previous modification that allowed dict tokens.

### DIFF
--- a/src/client/client.js
+++ b/src/client/client.js
@@ -14,17 +14,21 @@ function removeEmpty(obj) {
     return obj;
 }
 
-function HTTPClient(requestHeaders, baseServer, port) {
+function HTTPClient(token, baseServer, port, headers={}) {
     // Do not need colon if port is empty
     if (port !== '') {
         baseServer += ":" + port.toString();
     }
     this.address = baseServer;
+    this.token = token;
+    this.clientHeaders = headers;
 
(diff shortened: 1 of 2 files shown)


pzbitskiy:
@fxgamundi please rebase/resolve conflicts.

pzbitskiy:
@fxgamundi it is hard to review because the diff includes some commits from master. Suggest rebasing on top master and force pushing to have only your 5 commits in the diff.

fxgamundi:
Sorry for that. Now everything should be clean!

src/client/client.js
```diff
@@ -14,17 +14,21 @@ function removeEmpty(obj) {
...
 
-function HTTPClient(requestHeaders, baseServer, port) {
+function HTTPClient(token, baseServer, port, headers={}) {
     // Do not need colon if port is empty
     if (port !== '') {
         baseServer += ":" + port.toString();
     }
     this.address = baseServer;
+    this.token = token;
+    this.clientHeaders = headers;
```
rotemh: Can we call it `DefaultHeaders`? Client seems a bit off to me in this context.

PR #50, opened 2019-09-13
Draft: asset creation

EvanJRichard:


mjiang102628:
The optional handling part looks good; I think you have to use Buffer.from for most of the addresses (manager, freeze, etc) since they need to be encoded as bytes (not string). Probably also do this for unitname and assetname since these should also be bytes (or convert to bytes if we want the constructor to take strings).

mjiang102628:
Also, some transaction parameters need to go under AssetParams; the json should look something like
```json
{
  "apar": {
    "c:b64": "CfvSdiwI+Gxa5r9t16epAd5mdddQ4H6MXHaYZH224f0=",
    "f:b64": "CfvSdiwI+Gxa5r9t16epAd5mdddQ4H6MXHaYZH224f0=",
    "m:b64": "CfvSdiwI+Gxa5r9t16epAd5mdddQ4H6MXHaYZH224f0=",
    "r:b64": "CfvSdiwI+Gxa5r9t16epAd5mdddQ4H6MXHaYZH224f0="
  },
  "fee": 3790,
  "fv": 322575,
  "gh:b64": "SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=",
  "lv": 323575,
  "snd:b64": "CfvSdiwI+Gxa5r9t16epAd5mdddQ4H6MXHaYZH224f0=",
  "type": "acfg"
}
```

EvanJRichard:
_@mjiang102628 wrote:_
> Also, some transaction parameters need to go under AssetParams; the json should look something like [...]

Cool, good find! I'm a little new to this sort of thing: does this mainly impact `get_obj_for_encoding` and `from_obj_for_encoding`?
Also, I think this may have been merged prematurely? I can revert the merge if so.

mjiang102628:
> Cool, good find! I'm a little new to this sort of thing: does this mainly impact `get_obj_for_encoding` and `from_obj_for_encoding`?
> Also, I think this may have been merged prematurely? I can revert the merge if so.
It says I merged something here, but I don't think I did (I only merged my branch); it doesn't show up on the list of commits, and the travis build isn't failing against a golden value for encoding/signing (which it should have if the encoding didn't agree with the go sdk). 

I believe it impacts both; since `get_obj_for_encoding` needs to put some parameters under "apar" and `from_obj_for_encoding` needs to retrieve those values.

PR #60, opened 2019-10-03
Add asset freeze/unfreeze txns.

EvanJRichard:


mjiang102628:
Add a couple of tests with golden values for encoding/decoding and make sure they work in another SDK as well. Otherwise, LGTM!

src/transaction.js
```diff
@@ -187,6 +188,32 @@ class Transaction {
...
+
+            };
+            if (this.creator !== undefined) txn.faid.c = Buffer.from(this.creator.publicKey);
+            if (this.freezeAccount !== undefined) txn.fadd = Buffer.from(this.freezeAccount.publicKey);
+            // allowed zero values
+            if (!txn.note.length) delete txn.note;
+            if (!txn.amt) delete txn.amt;
+            if (!txn.fee) delete txn.fee;
+            if (!txn.gen) delete txn.gen;
+            return txn;
```
pzbitskiy: it is probably ok for now but I think we need to refactor serialization

PR #64, opened 2019-10-11
LogicSig implementation

pzbitskiy:
Similar to all other SDKs - LogicSig creation, association with transactions, serialization/deserialization, program checks for length and cost.


src/logic/logic.js
```diff
@@ -0,0 +1,159 @@
+/**
+ Utilities for working with program bytes.
+ */
+
+const langspec = require("./langspec.json")
```
rotemh: We should probably checksum this file and compare upon loading to prevent erroneous changes to the file.

pzbitskiy: I think it is not a big deal - client side checking mainly aims reducing the amount of invalid programs send to the network.

pzbitskiy:
Rebased

Issue #69, opened 2019-10-17
Add wrappers to main.js and update README with examples


rotemh:

EvanJRichard:
plan: inspired by go-sdk, do something like
`MakePayTxn`, `MakeKeyRegTxn`, `MakeAssetConfigTxn`, `MakeAssetDestroyTxn`, (Create?), `MakeAssetTransferTxn`, `MakeAssetRevokeTxn`, `MakeAssetFreezeTxn` and add examples to `README`

- [ ] `MakePayTxn`
- [ ] `MakeKeyRegTxn`
- [ ] `MakeAssetConfigTxn`
- [ ] `MakeAssetDestroyTxn`
- [ ] `MakeAssetTransferTxn`
- [ ] `MakeAssetRevokeTxn` 
- [ ] `MakeAssetFreezeTxn`
- [ ] `README.md`

PR #74, opened 2019-10-22
Add wrappers to main.js - functions for making various types of txn - and usage in README.md

EvanJRichard:
## Summary
Previous asset PRs added support for assets to class `Transaction`, but a user would have to read the code and tests to see how assets work. This PR proposes to add `make[Foo]Txn` helpers and similar to make transaction creation easier. It also proposes to update the `README.md` with usage.

## See also
https://github.com/algorand/js-algorand-sdk/issues/69

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,5 +307,141 @@ let algodclient = new algosdk.Algod(atoken, aserver, aport);
 algodclient.sendRawTransaction(rawSignedTxn);
 ```
 
+#### Assets
+
+Asset creation:
+```javascript
+let addr = "BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4";
+let fee = 10;
+let defaultFrozen = false;
+let genesisHash = "SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=";
+let totalIssuance = 100;
+let reserve = addr;
+let freeze = addr;
+let clawback = addr;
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -307,5 +307,141 @@ let algodclient = new algosdk.Algod(atoken, aserver, aport);
 algodclient.sendRawTransaction(rawSignedTxn);
 ```
 
+#### Assets
```
rotemh: Consider adding some words. Maybe a a few lines at the beginning. It's a bit out of context.

src/main.js
```diff
@@ -335,6 +335,263 @@ function signLogicSigTransaction(txn, lsig) {
     };
 }
 
+
+/**
+ * makePaymentTxn takes payment arguments and returns a transaction.Transaction object representing that payment txn
```
rotemh: I think a `Transaction` object is sufficient, no need for `transaction.Transaction`

tests/5.Transaction.js
```diff
@@ -333,4 +335,238 @@ describe('Sign', function () {
 
         });
     });
+
+    describe('transaction making functions', function () {
+        it('should be able to use helper to make a payment transaction', function() {
```
rotemh: It would better if these tests would compare to golden binaries. but it can be sufficient if `algosdk.makePaymentTxn` does it already.

EvanJRichard: :+1: Asset transaciton goldens are in `7.AlgoSDK.js`

PR #77, opened 2019-10-31
Added optional headers parameter to assetInformation

fxgamundi:
To keep the functionality on par with all the other methods


src/client/algod.js
```diff
@@ -198,10 +198,11 @@ function Algod(token = '', baseServer = "http://r2.algorand.network", port = 418
      * assetInformation returns the information for the asset with the passed creator and index
      * @param creator string
```
EvanJRichard: `Creator` is no longer needed :+1:

fxgamundi:
Merge fixed! Thanks

PR #78, opened 2019-11-01
Better handling of Transaction metadata hash

EvanJRichard:
Previous version of transaction `from_obj_for_encoding` just directly assigned `metadataHash`, rather than allocating a `new Uint8Array` (like Note).

While I was at it, I added length checking to the constructor, and removed a debug print I noticed in testing.

### Testing
Leaning on automated regression testing. Manually ran tests myself as well, goldens are untouched.


EvanJRichard:
Closing - I think there may be a small quality of life bug relating to this discussed change but I need to understand it better and prove it with test cases before requesting review

PR #82, opened 2019-11-12
Add lease support.

EvanJRichard:
## Summary
Add support for leases. 

### Testing
Added `golden` test to `7.AlgoSDK.js`

### See also
https://github.com/algorand/js-algorand-sdk/issues/80
https://github.com/algorand/go-algorand-sdk/pull/80


src/main.js
```diff
@@ -566,10 +582,11 @@ function makeAssetFreezeTxn(from, fee, firstRound, lastRound, note, genesisHash,
  * @param genesisHash - string specifies hash genesis block of network in use
  * @param genesisID - string specifies genesis ID of network in use
  * @param assetIndex - int asset index uniquely specifying the asset
+ * @param lease - see makePaymentTxn
  * @returns {Transaction}
  */
 function makeAssetTransferTxn(from, to, closeRemainderTo, revocationTarget,
-                              fee, amount, firstRound, lastRound, note, genesisHash, genesisID, assetIndex) {
+                              fee, amount, firstRound, lastRound, note, genesisHash, genesisID, assetIndex, lease) {
```
rotemh: No need for that given the `addLease` method

rotemh: Same for the rest

tests/7.AlgoSDK.js
```diff
@@ -83,6 +83,33 @@ describe('Algosdk (AKA end to end)', function () {
             let tx_golden = "5FJDJD5LMZC3EHUYYJNH5I23U4X6H2KXABNDGPIL557ZMJ33GZHQ";
             assert.deepStrictEqual(js_dec.txID, tx_golden);
         });
+
+        it('should return a blob that matches the go code when adding a lease', function () {
```
rotemh: It would be better the create the transaction using the `MakePaymentTransaction` and `txn.addLease` functions rather than generating the object yourself.

src/main.js
```diff
@@ -347,9 +347,14 @@ function signLogicSigTransaction(txn, lsig) {
...
  * @param genesisID - string specifies genesis ID of network in use
+ * @param lease - 32 byte array or undefined - Lease enforces mutual exclusion of transactions.  If this field is
+ * nonzero, then once the transaction is confirmed, it acquires the
+ * lease identified by the (Sender, Lease) pair of the transaction until
+ * the LastValid round passes.  While this transaction possesses the
+ * lease, no other transaction specifying this lease can be confirmed.
  * @returns {Transaction}
  */
-function makePaymentTxn(from, to, fee, amount, firstRound, lastRound, note, genesisHash, genesisID) {
+function makePaymentTxn(from, to, fee, amount, firstRound, lastRound, note, genesisHash, genesisID, lease) {
```
rotemh: No need for that given the addLease method

src/transaction.js
```diff
@@ -45,6 +46,13 @@ class Transaction {
         else {
           note = new Uint8Array(0);
         }
+        if (lease !== undefined) {
+            if (lease.constructor !== Uint8Array) throw Error("lease must be a Uint8Array.");
+            if (lease.length !== ALGORAND_TRANSACTION_LEASE_LENGTH) throw Error("lease must be of length " + ALGORAND_TRANSACTION_LEASE_LENGTH.toString() + ".");
+        }
+        else {
+            lease = new Uint8Array(0);
+        }
```
rotemh: It's a duplicate of the `addLease` function. remove it and keep just the other one.

EvanJRichard: @rotem I'm removing it from the helpers, but would like to keep it as an optional constructor arg like this.

A common way of building and signing transactions is building constructor arguments in a dict and then passing the constructors to `algosdk.signTransaction` to immediately get the signed transaction back - see example in the test `it('should return a blob that matches the go code when using a flat fee'`.

`AddLease` can't be called on a signed transaction, though, and I wanted to support the `algosdk.signTransaction` path.

rotemh: I'm ok with the constructor but the one thing we can't do is to break the API. This SDK update should be backwards compatible so we can't change exiting ones. One unfortunate result is that we will not be able to support the `algosdk.signTransaction` path. 

Bottom line - We can leave the constructor trick but we can't change any of the `main.js` functions.

EvanJRichard: Sounds good - current commit keeps the constructor trick but leaves `main.js` as is

PR #86, opened 2019-11-19
LogicSig templates: HTLC, split

EvanJRichard:
## Summary
For example usage, see new tests in `8.LogicSig.js`

### Tests
Golden testing in `8.LogicSig.js`

### DIFF
--- a/src/logicTemplates/split.js
+++ b/src/logicTemplates/split.js
@@ -1,3 +1,5 @@
+const templates = require('./templates')
+
 class Split {
     /**
     * Split splits money sent to some account to two recipients at some ratio.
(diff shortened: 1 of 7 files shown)


src/logicTemplates/templates.js
```diff
@@ -0,0 +1,75 @@
...
+                decodedLength = putUvarint(buf, val);
+                res = replace(res, buf, offsets[i], 1);
+                break;
+            case valTypes.ADDRESS:
+                val = address.decode(val);
+                res = replace(res, val.publicKey, offsets[i], 32);
+                break;
+            case valTypes.BASE64:
+                val = Buffer.from(val, 'base64');
+                res = replace(res, val, offsets[i], 32);
```
rotemh: I'd add a comment that the place holder of the b64 is always 32

src/logicTemplates/templates.js
```diff
@@ -0,0 +1,75 @@
...
+
+    return res
+}
+
+function addressFromProgram(program) {
+    let tag = Buffer.from("Program");
+    let forHashing = Buffer.concat([tag, program]);
+    let hash = nacl.genericHash(forHashing);
+    return address.encode(hash);
+}
```
rotemh: Doesn't logic have this function?

src/main.js
```diff
@@ -347,9 +347,14 @@ function signLogicSigTransaction(txn, lsig) {
...
  * @param genesisID - string specifies genesis ID of network in use
+ * @param lease - 32 byte array or undefined - Lease enforces mutual exclusion of transactions.  If this field is
+ * nonzero, then once the transaction is confirmed, it acquires the
+ * lease identified by the (Sender, Lease) pair of the transaction until
+ * the LastValid round passes.  While this transaction possesses the
+ * lease, no other transaction specifying this lease can be confirmed.
  * @returns {Transaction}
  */
-function makePaymentTxn(from, to, fee, amount, firstRound, lastRound, note, genesisHash, genesisID) {
+function makePaymentTxn(from, to, fee, amount, firstRound, lastRound, note, genesisHash, genesisID, lease) {
```
rotemh: I think that this line sneaked in from another PR. (which we also changed)

src/main.js
```diff
@@ -381,10 +387,11 @@ function makePaymentTxn(from, to, fee, amount, firstRound, lastRound, note, gene
  * @param voteFirst - first round on which voteKey is valid
  * @param voteLast - last round on which voteKey is valid
  * @param voteKeyDilution - integer
+ * @param lease - see makePaymentTxn
  * @returns {Transaction}
  */
 function makeKeyRegistrationTxn(from, fee, firstRound, lastRound, note, genesisHash, genesisID,
-                                voteKey, selectionKey, voteFirst, voteLast, voteKeyDilution) {
+                                voteKey, selectionKey, voteFirst, voteLast, voteKeyDilution, lease) {
```
rotemh: Same here and for the rest.

src/transaction.js
```diff
@@ -45,6 +46,13 @@ class Transaction {
         else {
           note = new Uint8Array(0);
         }
+        if (lease !== undefined) {
+            if (lease.constructor !== Uint8Array) throw Error("lease must be a Uint8Array.");
+            if (lease.length !== ALGORAND_TRANSACTION_LEASE_LENGTH) throw Error("lease must be of length " + ALGORAND_TRANSACTION_LEASE_LENGTH.toString() + ".");
+        }
+        else {
+            lease = new Uint8Array(0);
+        }
```
rotemh: I think the PRs got mixed.

src/logicTemplates/split.js
```diff
@@ -0,0 +1,121 @@
...
+     * @param {string} genesisHash: the b64-encoded genesis hash indicating the network for this transaction
+     * @param {boolean} precise, optional, precise treats the case where amount is not perfectly divisible based on the ratio.
+     *  When set to False, the amount will be divided as close as possible but one address will get
+     *  slightly more. When True, an error will be raised.
+     * @returns {Uint8Array}
+     */
+    getSendFundsTransaction(amount, firstRound, lastRound, fee, genesisHash, precise=true) {
+        let amountForReceiverOne = 0;
+        // reduce fractions
+        var gcdFn = function(a, b) {
```
rotemh: may be good to check that these are numbers 
```javascript
if ((typeof a !== 'number') || (typeof b !== 'number')) throw "gcd operates only on positive integers";
```

or something along this line.

src/logicTemplates/split.js
```diff
@@ -0,0 +1,121 @@
...
+            throw Error("precise splitting requested but amount and contract ratio cannot be split precisely");
+        } else {
+            amountForReceiverOne = Math.round(amount * ratn / ratd);
+        }
+        let amountForReceiverTwo = amount - amountForReceiverOne;
+
+        let from = this.address;
+
+        let tx1 = algosdk.makePaymentTxn(from, this.receiverOne, fee, amountForReceiverOne, firstRound, lastRound, undefined, genesisHash, undefined, undefined);
+        let tx2 = algosdk.makePaymentTxn(from, this.receiverTwo, fee, amountForReceiverTwo, firstRound, lastRound, undefined, genesisHash, undefined, undefined);
```
rotemh: I think that these two lines don't take into account the change to the interface (no lease)

src/logicTemplates/templates.js
```diff
@@ -0,0 +1,75 @@
...
+};
+
+function inject(orig, offsets, values, valueTypes) {
+    assert.strictEqual(offsets.length, values.length);
+    assert.strictEqual(offsets.length, valueTypes.length);
+
+    let res = orig;
+
+    function replace(arr, newVal, offset, placeholderLength) {
+        let beforeReplacement = arr.slice(0, offset); // TODO ejr maybe offset + 1
```
rotemh: is it a leftover comment?

src/logicTemplates/templates.js
rotemh: Also, you should take into account the length of the b64 value. it's not always going to be 32 bytes long.

Issue #100, opened 2019-12-05
Transaction Group support for LogicSig signed transactions


k3karthic-bleumi:
Making a transaction group where one or more of the transactions are signed by a LogicSig does not work. 

The 'signLogicSigTransaction' function internally creates a transaction object from a transaction dictionary and returns the txID & encoded blob. However, the Transaction class constructor does not support passing the groupID as a parameter and sets the internal group parameter to undefined as can be seen in the code below.

https://github.com/algorand/js-algorand-sdk/blob/3ba7463223aab24f67cffb30748679502a4f8d1a/src/transaction.js#L82

This causes the encoded transaction to fail with the error 'empty txgroup' when submitted to the network.

To workaround this issue, we import the internal 'src/transaction.js' & 'src/encoding/encoding.js' modules and reimplement the 'signLogicSigTransaction' function to ensure the groupID is set on the transaction object.

EvanJRichard:
PRs are in progress to fix! Thanks a million again for noting this. :+1:

PR #111, opened 2020-01-17
Fix dynamicFee and transaction group encoding

### LINKED ISSUE #108: Transaction Group support for ASA Transfers
The _'get_obj_for_encoding'_ method on the Transaction class does not seem to set the _'grp'_ parameter for ASA transfers. This causes a transaction group to fail with the error message _'TransactionPool.Remember: transactionGroup: [0] had zero Group but was submitted in a group of 2'_.

I had to monkey patch the method to get it to work,
```
for (let mp of txns) {
    let orig = mp.get_obj_for_encoding;
    let replace = function() {
        let obj = orig.bind(this)();
        obj.grp = this.group;
        return obj;
    };
    mp.get_obj_for_encoding = replace.bind(mp);
}
```

### LINKED ISSUE #110: Fix dynamic fee contract

EvanJRichard:
## Summary
This PR fixes dynamicFee, which was producing incorrect `stxn`s. While getting that working, a bug relating to transaction groups was fixed.

### Testing
Golden testing.

### See also
This fixes #108 but not #100 
This closes #110


src/logicTemplates/dynamicfee.js
```diff
@@ -0,0 +1,180 @@
...
+ * @param {dict} txn - main transaction from payer's signDynamicFee output (a dict of constructor arguments, NOT a transaction.Transaction)
+ * @param {LogicSig} lsig - the signed logic received from the payer's signDynamicFee output
+ * @param {Uint8Array} privateKey - the private key for the account that pays the fee
+ * @param {int} fee - fee per byte for both transactions
+ * @param {int} firstValid - first protocol round on which both transactions will be valid
+ * @param {int} lastValid - last protocol round on which both transactions will be valid
+ *
+ * @throws on invalid lsig
+ */
+function getDynamicFeeTransactions (txn, lsig, privateKey, fee, firstValid, lastValid) {
```
winder: Should this be `getReimbursementTransaction`? Maybe I named the function wrong in java.

src/logicTemplates/dynamicfee.js
```diff
@@ -0,0 +1,180 @@
...
+    if (!lsig.verify(address.decode(txn.from).publicKey)) {
+        throw new Error("invalid signature");
+    }
+
+    txn.firstRound = firstValid;
+    txn.lastRound = lastValid;
+    txn.fee = fee;
```
winder: This is fee per byte, so you'll need `txn.fee = fee * estimatedSize(txn)`

src/logicTemplates/dynamicfee.js
```diff
@@ -0,0 +1,180 @@
...
+    if (txn.fee < transaction.ALGORAND_MIN_TX_FEE) {
+        txn.fee = transaction.ALGORAND_MIN_TX_FEE
+    }
+
+    let keys = nacl.keyPairFromSecretKey(privateKey);
+    let from = address.encode(keys.publicKey);
+
+    // must remove lease and re-add using addLease so that fee calculation will match other SDKs
```
winder: You should be able to remove all the `txnObj` stuff and just use `txn` after fixing the calculation above.

src/logicTemplates/dynamicfee.js
EvanJRichard: could be either, I think both are accurate, submitting the resultant bytes both sends a transaction with a dynamic fee and a transaction reimbursing the fee

src/logicTemplates/dynamicfee.js
EvanJRichard: `estimateSize` is a method on an already-constructed transaction, so letting the constructor handle the feePerByte conversion seems to be the only way to end up with the same fee as in the other SDKs.

Another example of doing this: a transaction can be constructed with `lease` from the outset, but using `addLease` is necessary to replicate the same fee output.

src/logicTemplates/dynamicfee.js
```diff
@@ -0,0 +1,180 @@
...
+        if (!Number.isSafeInteger(firstValid) || firstValid < 0) throw Error("firstValid must be a positive number and smaller than 2^53-1");
+        if (!Number.isSafeInteger(lastValid) || lastValid < 0) throw Error("lastValid must be a positive number and smaller than 2^53-1");
+
+        if (closeRemainder === undefined) {
+            closeRemainder = this.zeroAddress
+        }
+        if (lease === undefined) {
+            let leaseBytes = nacl.randomBytes(32);
+            lease = Buffer.from(leaseBytes).toString('base64');
+        }
```
btoll: Do you mean to explicitly test for only `undefined` here?  Could you do `if (!closeRemainder)` or even set a default value in the constructor signature (where it's readable, of course)?

These are nits :)

src/logicTemplates/dynamicfee.js
```diff
@@ -0,0 +1,180 @@
+const templates = require('./templates');
```
pzbitskiy: We already have this file in master, what is the difference?

PR #112, opened 2020-01-17
Extra functions in algosdk

### LINKED ISSUE #100: Transaction Group support for LogicSig signed transactions

### LINKED ISSUE #101: Expose LogicSig.fromByte
We can create and serialize a LogicSig object using the 'makeLogicSig' method,

```
// create LogicSig object and sign with our secret key
let program = Uint8Array.from([1, 32, 1, 0, 34]);  // int 0 => never transfer money
let lsig = algosdk.makeLogicSig(program);
lsig.sign(keys.secretKey);

assert lsig.verify(sender_pk);

console.log(lsig.toByte()); // Serialized LogicSig
```

However, the deserialization method 'fromByte' is declared as a static method as can be seen in the code below.

https://github.com/algorand/js-algorand-sdk/blob/3ba7463223aab24f67cffb30748679502a4f8d1a/src/logicsig.js#L150

As the LogicSig class is not exposed on the algosdk module, we had to import the internal 'src/logicsig.js' module. Could we have this class or the fromByte method exposed on the algosdk object?

### LINKED ISSUE #109: Transaction group support for all kinds of transaction
See also #100 #108 #70

EvanJRichard:
## Summary
Add new functions in `algosdk` that expose internal functions developers need, or that accept slightly different arguments:
- `logicSigFromByte` takes encoded bytes to produce a logicsig
- `signLogicSigTransactionObject` is as `signLogicSigTransaction`, but it expects already-constructed `Transaction` rather than a dict of constructor arguments. This allows the use of groupID in combination with logicsigs.

### See also
Resolves #100 `signLogicSigTransactionObject`
Resolves #101 `logicSigFromByte`
Closes #109 out

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -335,6 +335,33 @@ function signLogicSigTransaction(txn, lsig) {
     };
 }
 
+/**
+ * signLogicSigTransactionObject takes transaction.Transaction and a LogicSig object and returns a logicsig
+ * transaction which is a blob representing a transaction and logicsig object.
+ * @param {Object} txn transaction.Transaction
+ * @param {LogicSig} lsig logicsig object
+ * @returns {Object} Object containing txID and blob representing signed transaction.
+ */
+function signLogicSigTransactionObject(txn, lsig) {
+    let lstx = {
+        lsig: lsig.get_obj_for_encoding(),
+        txn: txn.get_obj_for_encoding()
+    };
+
+    return {
+        "txID": txn.txID().toString(),
+        "blob": encoding.encode(lstx)
+    };
+}
+
+/**
+ * logicSigFromByte accepts encoded logic sig bytes and attempts to call logicsig.fromByte on it,
+ * returning the result
+ */
+function logicSigFromByte(encoded) {
+    return logicsig.LogicSig.fromByte(encoded);
+}
+
 
 /**
  * makePaymentTxn takes payment arguments and returns a Transaction object
@@ -624,6 +651,8 @@ module.exports = {
     assignGroupID,
     makeLogicSig,
     signLogicSigTransaction,
+    signLogicSigTransactionObject,
+    logicSigFromByte,
     makePaymentTxn,
     makeKeyRegistrationTxn,
     makeAssetCreateTxn,
--- a/src/main.js
+++ b/src/main.js
@@ -312,7 +312,7 @@ function makeLogicSig(program, args) {
 /**
  * signLogicSigTransaction takes  a raw transaction and a LogicSig object and returns a logicsig
  * transaction which is a blob representing a transaction and logicsig object.
- * @param {Object} txn transaction object
+ * @param {Dict} txn transaction constructors dictionary
  * @param {LogicSig} lsig logicsig object
  * @returns {Object} Object containing txID and blob representing signed transaction.
  * @throws error on failure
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -335,6 +335,33 @@ function signLogicSigTransaction(txn, lsig) {
...
+    };
+}
+
+/**
+ * logicSigFromByte accepts encoded logic sig bytes and attempts to call logicsig.fromByte on it,
+ * returning the result
+ */
+function logicSigFromByte(encoded) {
+    return logicsig.LogicSig.fromByte(encoded);
+}
```
btoll: Could `logicsig.LogicSig` ever be null?  Should you guard against it?

src/main.js
```diff
@@ -312,7 +312,7 @@ function makeLogicSig(program, args) {
 /**
  * signLogicSigTransaction takes  a raw transaction and a LogicSig object and returns a logicsig
  * transaction which is a blob representing a transaction and logicsig object.
- * @param {Object} txn transaction object
+ * @param {Dict} txn transaction constructors dictionary
```
pzbitskiy: JSDoc does not have Dict type

src/main.js
pzbitskiy: Nope, it's class in logicsig file

src/main.js
```diff
@@ -335,6 +335,33 @@ function signLogicSigTransaction(txn, lsig) {
...
+        "txID": txn.txID().toString(),
+        "blob": encoding.encode(lstx)
+    };
+}
+
+/**
+ * logicSigFromByte accepts encoded logic sig bytes and attempts to call logicsig.fromByte on it,
+ * returning the result
+ */
+function logicSigFromByte(encoded) {
```
pzbitskiy: do we really need this thin wrapper?

src/main.js
```diff
@@ -335,6 +335,33 @@ function signLogicSigTransaction(txn, lsig) {
...
 }
 
+/**
+ * signLogicSigTransactionObject takes transaction.Transaction and a LogicSig object and returns a logicsig
+ * transaction which is a blob representing a transaction and logicsig object.
+ * @param {Object} txn transaction.Transaction
+ * @param {LogicSig} lsig logicsig object
+ * @returns {Object} Object containing txID and blob representing signed transaction.
+ */
+function signLogicSigTransactionObject(txn, lsig) {
```
pzbitskiy: test for it?

EvanJRichard: Deduplicated code so it's now included in the test

src/main.js
EvanJRichard: Yes, without this thin wrapper users have the ability to serialize but not deserialize logicsigs - see https://github.com/algorand/js-algorand-sdk/issues/101

PR #120, opened 2020-03-09
Transaction accepts params object

EvanJRichard:
## Summary
Class `Transaction` accepts a `params` object, and preferentially uses that to set up `lastRound` `genesisHash` et al. 
`algosdk.MakeFooTxn` helpers now have counterparts in the form of `MakeFooTxnWithParams`. `MakeFooTxn` marked `@Deprecated`, uses `params`-based implementation internally.
Add `suggestParams` function to `algod.js`, which uses `getTransactionParams` to build the `params` object that `Transaction` expects.

### Testing
Unit tests pass; known broken cucumber tests fail. Because the old `MakeFoo` family of funcs calls the new `MakeFoo` family of funcs, they are tested by existing unit tests.

### See also
https://github.com/algorand/js-algorand-sdk/issues/105

### DIFF
--- a/src/client/algod.js
+++ b/src/client/algod.js
@@ -249,6 +249,23 @@ function Algod(token = '', baseServer = "http://r2.algorand.network", port = 418
     };
 
     /**
+     * suggestParams returns to common needed parameters for a new transaction, in a format the transaction builder expects
+     * @param headers, optional
+     * @returns {Object}
+     */
+    this.suggestParams = async function (headers={}) {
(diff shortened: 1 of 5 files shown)


EvanJRichard:
_@mjiang102628 wrote:_
> Can you add some tests for the new makeTxnWithParams functions?
The unit tests currently test these functions: they call the old `makeTxn` functions, which have been changed to call the new `makeTxnWithParams`.

mjiang102628:
Oh, my bad. You're right.

PR #151, opened 2020-05-10
Update Keccak library to 3.0.0

### LINKED ISSUE #126: Does not install with latest Node 12 LTS
Hi,

This release added a dependency on `keccak256` which in turn uses an outdated version of the native `keccak` library (1.4.0) causing node-gyp to fail compilation. 

The native version v3.0.0 has this issue fixed but the middleware library does not use it.

Regards,
Mauro.

jecassis:
The `keccak256` wrapper package is not needed as the only usage in the SDK is already guaranteed to be a buffer. So dropped it and in the process updated the `keccak` package to 3.0.0 and made it a direct dependency to fix the failure in #126 where the function overloads in the C++ library without arguments where removed from Node.js's V8 bindings.

Fixes #126


src/logicTemplates/htlc.js
```diff
@@ -98,7 +98,7 @@ function signTransactionWithHTLCUnlock(contract, txn, preImageAsBase64) {
             throw new Error("sha256 hash of preimage did not match stored contract hash")
         }
     } else if (hashFunction == 2) {
-        let actualHashedOutput = keccak256(preImageBytes);
+        let actualHashedOutput = createKeccakHash('keccak256').update(preImageBytes).digest();
```
algorandskiy, line 101: imported `keccak256` but calling `createKeccakHash` this should be working.

jecassis: You're right. Forgot to commit that hunk, I've updated the pull request with the change and rebased it against `develop`.

algorandskiy:
All right, waiting resolution about the test for `hashFunction == 2` case.

ryanRfox:
I am cherry picking this with success on node 12.18.0

jasonpaulos:
@algorandskiy I have added a test for `hashFunction == 2`.

algorandskiy (approved):
Thank you Jason. Let's rebase and merge

PR #155, opened 2020-05-13
Rekeying transactions

EvanJRichard:
## Summary
An upcoming protocol change will allow accounts to change the spending key associated with the public address. This PR proposes to support this feature:
- `Transaction`s can be created with a `reKeyTo: addr` field in the constructor
- Alternatively, `Transaction`s support `addRekey`, like adding a lease
- `algosdk.signTransaction` now only writes the `txn.from` field if none was supplied
- `stepdefs` added per `algorand-sdk-testing:rotem/rekey`
- `signTxn` and `signTransaction` will add the `sgnr` authorizing address field if the supplied `sk` doesn't match `txn.from.pk`

### Testing
All tests pass in local Docker container. _Note_, `run_docker.sh` has been left on `develop`


src/transaction.js
```diff
@@ -412,6 +431,15 @@ class Transaction {
             this.fee += 37 * feePerByte; // 32 bytes + 5 byte label
         }
     }
+
+    // add the rekey-to field to a transaction not yet having it
+    // supply feePerByte to increment fee accordingly
+    addRekey(reKeyTo, feePerByte=0) {
```
rotemh, line 437: why do we add `feePerByte` here? Isn't sizeEstimate happening during the signing process?

EvanJRichard: The final `fee` is calculated at construction, not at signing, so if you mutate the `Transaction` after construction, this function provides the option to update the fee - like we did with `addLease`, above.

src/transaction.js
```diff
@@ -412,6 +431,15 @@ class Transaction {
...
+        if (reKeyTo !== undefined) this.reKeyTo = address.decode(reKeyTo);
+        if (feePerByte !== 0) {
+            this.fee += 37 * feePerByte; // 32 bytes + 5 byte label
```
rotemh, line 440: consider making 37 a constant.

PR #158, opened 2020-05-14
Indexer healthcheck endpoint

winder:


src/client/v2/indexer/makeHealthCheck.js
```diff
@@ -0,0 +1,17 @@
+class MakeHealthCheck{
+	constructor(c) {
+		this.c = c;
+	}
+
+	/**
+	 * returns the block for the passed round
```
EvanJRichard, line 7: If you update this comment, then it LGTM :+1:

PR #177, opened 2020-06-26
Added helper function to correctly set Accept on GETs

fxgamundi:
Added helper function that can be easily expanded for additional responses formats if needed.

### DIFF
--- a/src/client/client.js
+++ b/src/client/client.js
@@ -14,6 +14,26 @@ function removeEmpty(obj) {
     return obj;
 }
 
+/**
+ * getAccceptFormat returns the correct Accept header depending on the
+ * requested format.
+ * @param query
+ * @returns {string}
+ */
+function getAccceptFormat(query) {
+    if('format' in query) 
+        switch(query['format']) {
+            case 'msgpack':
+                return 'application/msgpack';
+            case 'json':
+                return "application/json"
+            default:
+                return "application/json"
+        }
+    else
+        return "application/json"
+ }
+
 function HTTPClient(token, baseServer, port, headers={}) {
     // Do not need colon if port is empty
     if (port !== '') {
@@ -30,7 +50,7 @@ function HTTPClient(token, baseServer, port, headers={}) {
                 .set(this.token)
                 .set(this.defaultHeaders)
                 .set(requestHeaders)
-                .set('Accept', 'application/json')
+                .set('Accept', getAccceptFormat(query))
                 .query(removeEmpty(query));
         } catch (e) {
             throw e;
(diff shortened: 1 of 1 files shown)


fxgamundi:
Pushed a commit that addresses Travis' previous error, but now its failing on `docker-test`

src/client/client.js
```diff
@@ -14,6 +14,26 @@ function removeEmpty(obj) {
...
 }
 
+/**
+ * getAccceptFormat returns the correct Accept header depending on the
+ * requested format.
+ * @param query
+ * @returns {string}
+ */
+function getAccceptFormat(query) {
+    if (typeof query !== "undefined" && query.hasOwnProperty('format'))
```
algorandskiy, line 24: please be consistent with quotes, here are both single and double quotes in the same like. Plus, it this context there is no difference between `typeof query !== "undefined"` and `query !== undefined` since it is declared.

src/client/client.js
```diff
@@ -14,6 +14,26 @@ function removeEmpty(obj) {
...
+        switch(query['format']) {
```
algorandskiy, line 25: `query.format` looks more consistent.

algorandskiy:
@algorotem do we still accept fixes for client v1?

algorotem:
> @algorotem do we still accept fixes for client v1?

No, just critical bugs.

src/client/client.js
fxgamundi: You're totally right. Changed them in the latest commit

src/client/client.js
fxgamundi: Also changed in latest commit

fxgamundi:
This is not a fix for v1, this is a fix for the HTTP client that is used in both v2 and indexer clients.

src/client/client.js
```diff
@@ -14,6 +14,26 @@ function removeEmpty(obj) {
...
+            case 'msgpack':
+                return 'application/msgpack';
+            case 'json':
+                return 'application/json';
+            default:
+                return 'application/json';
+        }
+    else
+        return "application/json"
+ }
```
algorotem, line 35: tiny comment - since we're dealing with only two type, this would be sufficient.
```js
function getAccceptFormat(query) {
	if (query !== undefined && query.hasOwnProperty('format'))
		if (query.format === 'msgpack')
			return 'application/msgpack';
	return "application/json"
 }
```

fxgamundi: We thought of that approach, but wanted to present a solution that allowed to easily add additional types, thinking that in the future that might happen. But if there's only going to be two types then I'll change the code to yours!

PR #187, opened 2020-07-11
Response models for api v2

algorandskiy:


algorandskiy:
Yeah, I think this code can be used as a template for a generator. we'll see.

src/client/v2/algod/models/base.js
```diff
@@ -48,6 +48,46 @@ class BaseModel {
         }
         return obj;
     }
+
+    static from_obj_for_encoding(obj) {
+        const target = {};
+        for (const key of Object.keys(obj)) {
+            if (this.openapi_types[key] === undefined) {
```
winder, line 55: I think it's better to be lenient about new fields appearing in the responses. If we add new fields in an otherwise backwards compatible way it shouldn't break existing apps. Can you just add the field without the extra formatting?

algorandskiy: I think there are two options:
1. Silently convert to Object and silently lose typing and possibly later catch bugs in serialization
2. Allow the caller decide what to do if the object can't be deserialized - maybe leave it as Object.

What Java/Go SDKs do in this case?

winder: I think ignoring the unknown fields is fine, thats what Java does

PR #203, opened 2020-08-14
Change algosdk.signMultisigTransaction to accept either a built Transaction or a dict of constructor args

### LINKED ISSUE #188: V2 Multisig is failing to sign transactions created with transaction builder
Code to reproduce is below.

```
const algosdk = require('algosdk');

const token = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
const server = "http://localhost";
const port = 4001;

const keypress = async() => {
    process.stdin.setRawMode(true)
    return new Promise(resolve => process.stdin.once('data', () => {
        process.stdin.setRawMode(false)
        resolve()
    }))
}

(async() => {
    //create an account
    var account1 = algosdk.generateAccount();
    console.log(account1.addr);
    //create an account
    var account2 = algosdk.generateAccount();
    console.log(account2.addr);
    //create an account
    var account3 = algosdk.generateAccount();
    console.log(account3.addr);

    //Setup teh parameters for the multisig account
    const mparams = {
        version: 1,
        threshold: 2,
        addrs: [
            account1.addr,
            account2.addr,
            account3.addr,
        ],
    };

    var multsigaddr = algosdk.multisigAddress(mparams);
    console.log("Multisig Address: " + multsigaddr);
    //Pause execution to allow using the dispenser on testnet to put tokens in account
    console.log('Make sure address above has tokens using the dispenser');
    await keypress();
    try {
        let algodclient = new algosdk.Algodv2(token, server, port);

        //Get the relevant params from the algod
        let params = await algodclient.getTransactionParams().do();
        //example of how to write an object into the notefield
        let note = algosdk.encodeObj("Hello World");
        //create a transaction
        let txn = algosdk.makePaymentTxnWithSuggestedParams(multsigaddr, "7GBK5IJCWFPRWENNUEZI3K4CSE5KDIRSR55KWTSDDOBH3E3JJCKGCSFDGQ", 2000000, undefined, note, params);        

        //Sign wiith first signature
        let rawSignedTxn = algosdk.signMultisigTransaction(txn, mparams, account1.sk).blob;
        //sign with second account
        let twosigs = algosdk.appendSignMultisigTransaction(rawSignedTxn, mparams, account2.sk).blob;
        //submit the transaction
        let tx = (await algodclient.sendRawTransaction(twosigs));
        console.log("Transaction : " + tx.txId);

    } catch (err) {
        console.log(err);
    }
})().then(process.exit)
```

EvanJRichard:
# Summary
Previously `signMultisigTransaction` would only accept a dict of constructor args describing a transaction. Now, it can handle both that or a `Transaction` object.

Also, sort returned keys in indexer testing, because this was causing Travis to fail.

## Testing
Relying on existing unit tests to detect regressions. Manually tested that this code produced a blob and txid without error. Have not tried to use this to move TestNet algos. 

## Resolves
#188


EvanJRichard:
@winder Yeah, exactly - previously, "dict of constructor args" was used everywhere. But, more and more functions and tools are instead returning a completed `transaction.Transaction`, which can't be used with functions that were originally constructed for the dict pattern. I'll add a unit test for both paths. :+1:

winder (approved):
LGTM

PR #210, opened 2020-09-21
Fix v2 msgpack endpoints for browsers

### LINKED ISSUE #209: [Found work] Msgpack responses are undefined in browsers
@JasonWeathersby found that he could do operations like v2 `getBlock` and `pendingTransactionInformation` in node, but in a webapp the response would come back as `undefined`. Further investigation by @jasonpaulos found that this was likely due to how the superagent works: https://visionmedia.github.io/superagent/#parsing-response-bodies . Suggested solution from @winder and @jasonpaulos is to move `encoding.decode(res.body)` (from `do()` for the operations in question) into one of the `request.parse[application/msgp]` handlers mentioned in the link.

jasonpaulos:
This properly exposes msgpack responses in the browser for v2 endpoints such as `/v2/transactions/pending`.

Closes #209.

### DIFF
--- a/src/client/client.js
+++ b/src/client/client.js
@@ -45,13 +45,20 @@ function HTTPClient(token, baseServer, port, headers={}) {
 
     this.get = async function (path, query, requestHeaders={}) {
         try {
-            return await request
+            const format = getAccceptFormat(query);
+            let r = request
                 .get(this.address + path)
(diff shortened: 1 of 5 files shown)


EvanJRichard (approved):
Looks good to me! :ship:

PR #229, opened 2020-10-13
Use msgpack fork that supports bigints. 

### LINKED ISSUE #130: JS max integer values
I'm writing a client validation library for sanitizing user input for this SDK and ran into an edge case that I believe is worth a mention here. 

When trying to set the maximum number of assets that you're allowed to issue to `18446744073709551616`, javascript hits its max integer value, which rounds and loses precision.

Basically,
```javascript
parseInt('18446744073709551616')  // returns 18446744073709552000
```

The traditional way I understand to get around this is to convert to `BigInt` instead, but if this SDK is intended for browser use it will also need to be packaged with a polyfill to handle some less capable browsers. If this SDK is also using the standard Number type it could cause downstream issues from a conversion like this. 

Happy to help how I can, but i'd love your feedback to see if i'm missing something obvious.

EvanJRichard:
## Summary
See #130. This PR proposes to change the js-sdk to use a fork that supports `BigInt`<->`uint64` encode/decode. 

### See also
Resolves #130

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1192,12 +1192,9 @@
       }
     },
     "@msgpack/msgpack": {
-      "version": "1.9.0",
-      "resolved": "https://registry.npmjs.org/@msgpack/msgpack/-/msgpack-1.9.0.tgz",
-      "integrity": "sha512-m3Z/dOzxgpHMoyeYInpQEPti/ThRRHdtU1ZtHZtJZhoRv9J+xRrRDEUaOf5c2GFf26jlc9HIqF2/ubeI46cvqQ==",
-      "requires": {
-        "base64-js": "^1.3.0"
-      }
+      "version": "1.12.2",
+      "resolved": "https://registry.npmjs.org/@msgpack/msgpack/-/msgpack-1.12.2.tgz",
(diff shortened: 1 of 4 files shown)


jasonpaulos (commented):
@EvanJRichard could you add a small test to `tests/2.Encoding.js` that verifies encoding/decoding a BigInt works as expected?

Pending that, LGTM.

jasonpaulos (approved):

mxmauro:
Hi @EvanJRichard if I followed the correct path, `algo-msgpack-with-bigint` at NPM points to https://github.com/msgpack/msgpack-javascript 

That code encodes bigint as an extension and this breaks compatibility because uint64 is part of msgpack and Golang code and, almost sure, third party libraries, uses the uint64 format.

mxmauro:
Sorry @EvanJRichard my mistake, now found the correct code but still couldn't find the repo on GitHub

jasonpaulos:
@mxmauro the repo for `algo-msgpack-with-bigint` is https://github.com/EvanJRichard/msgpack-javascript.

@EvanJRichard, we should probably update our fork's package.json to point to its repo.

PR #231, opened 2020-10-14
Document that fee will be bumped up to min if finalFee < minFee

### LINKED ISSUE #224: Don't silently change the fee
Currently the transaction constructor silent changes the few when it is less than the minimal txn fee. This silently causes transactions to be authorized for more than is expected. It would be better for the transaction to be submitted with the lower fee and then rejected, so we could surface the error to the approver.

https://github.com/algorand/js-algorand-sdk/blob/c789ba84ab4de299544219fc7f5a13b23049dc1c/src/transaction.js#L133

This would be alleviate if the suggested parameters api call included an accurate fee value, but it doesn't.

EvanJRichard:
# Summary
When `Transaction` building is done, the `Transaction` class checks whether the final fee is below the protocol minimum fee. If it is, the class silently bumps the fee up to the minimum. This surprises user-devs. This PR proposes to add `If the final calculated fee is lower than the protocol minimum fee, the fee will be increased to match the minimum.` to all `main.js` functions involving the `fee`.

## Testing
Docs-only change; leaning on automated testing.

## See also
Closes #224

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -114,6 +114,8 @@ function masterDerivationKeyToMnemonic(mdk) {
  * 
  * Key registration fields: fee, firstRound, lastRound, voteKey, selectionKey, voteFirst,
  * voteLast, voteKeyDilution, genesisHash, note(optional), GenesisID(optional)
+ *
+ * If the final calculated fee is lower than the protocol minimum fee, the fee will be increased to match the minimum.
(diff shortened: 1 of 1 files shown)


jasonpaulos (approved):
It would be great to have this in the docs.

PR #233, opened 2020-10-22
Fixes for community tickets

### LINKED ISSUE #153: Allow `algosdk.signTransaction` to sign transactions generated by `make*`
Currently, if you use `algosdk.signTransaction` on a transaction generated using `algosdk. makeAssetTransferTxn`, it fails with the following error message:
```
```
Instead, you need to use the method `.signTxn`.

It would be great if `algosdk.signTransaction` behave as `.signTxn` for `Transaction` objects.

See https://forum.algorand.org/t/asset-transfer-by-grouping-transactions/1648/4

### LINKED ISSUE #220: Check program arguments are byte strings.
LogicSig ignores the args value and passes it silently through to encoding.encode, where it gets converted in msgpack bytes. But, goal does not decode these, instead treating them as byte string. I think that LogicSig should do a simple check that `args` is an array and that each element is from some basic set of safe things, such as strings, an arrays of numbers, or a Uint8array.

As it is, if you accidentally pass an object, then it will be converted by msgpack. It is plausible that this is useful, because maybe the logic program consumes msgpack bytes, but I think it is unlikely.

https://github.com/algorand/js-algorand-sdk/blob/3a052ea3109e212fb962c1691051dcc169f9343d/src/logicsig.js#L14

### LINKED ISSUE #222: Check SendRawTransaction arguments are valid
`SendRawTransaction` assumes its arguments are byte arrays, but does not check this. As a result, if you give it the result of `signLogicSigTransactionObject` rather than the `blob` property, then the logic signature is silently dropped from the transaction that gets submitted. This might only happen when you give it an array of txns (which is what I'm doing)

https://github.com/algorand/js-algorand-sdk/blob/fcfb0ced29c279afeac269c469971926cb504805/src/client/v2/algod/sendRawTransaction.js#L2

EvanJRichard:
Closes #153, closes #220, closes #222

### DIFF
--- a/src/logicsig.js
+++ b/src/logicsig.js
@@ -16,6 +16,14 @@ class LogicSig {
         this.tag = Buffer.from("Program");
 
         assert(logic.checkProgram(program, args));
+        if (args) {
+            assert(Array.isArray(args))
+            function checkType(arg) {
+                let theType = typeof arg;
+                return ((theType == "string") || (theType == "number") || (arg.constructor == Uint8Array));
+            }
+            assert(args.every(checkType))
+        }
 
         this.logic = program;
         this.args = args;
--- a/src/logicsig.js
+++ b/src/logicsig.js
@@ -20,7 +20,7 @@ class LogicSig {
             assert(Array.isArray(args))
             function checkType(arg) {
                 let theType = typeof arg;
-                return ((theType == "string") || (theType == "number") || (arg.constructor == Uint8Array) || (arg.constructor == Buffer));
+                return ((theType == "string") || (theType == "number") || (arg.constructor == Uint8Array) || (Buffer.isBuffer(arg)));
             }
             assert(args.every(checkType))
         }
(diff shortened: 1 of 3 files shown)


src/logicsig.js
```diff
@@ -16,6 +16,14 @@ class LogicSig {
         this.tag = Buffer.from("Program");
 
         assert(logic.checkProgram(program, args));
+        if (args) {
+            assert(Array.isArray(args))
+            function checkType(arg) {
+                let theType = typeof arg;
+                return ((theType == "string") || (theType == "number") || (arg.constructor == Uint8Array) || (arg.constructor == Buffer));
```
jasonpaulos, line 23: Nit: can you use `Buffer.isBuffer` to check for Buffers? That will work slightly better in browsers.

jasonpaulos (approved):
Looks good, I'm glad to more errors on invalid input!

jeapostrophe:
Thanks!

PR #235, opened 2020-10-26
Fix firefox tests

jasonpaulos:
* Switched to an Ubuntu base image for the docker test runner since installing Firefox on Debian proved to be more difficult.
* Firefox tests were failing because of the issue I pointed out here: https://github.com/mozilla/geckodriver/issues/1798. I implemented a workaround to this by using the helper functions `makeUint8Array` and `makeEmptyObject` and by fixing how we check for arrays in the SDK.
* I also found that the code for POST and DELETE requests throw a `Response` object instead of an `Error` if an error occurs, so I fixed this to improve the error message that gets displayed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ script:
   - make docker-test
 
 jobs:
-  allow_failures:
-    - name: Firefox test
   include:
     - stage: Testing
       name: Node test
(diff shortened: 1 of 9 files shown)


EvanJRichard (approved):
Nice, I like this. I think there was some reason for throwing `e.response` rather than `e`, but I can't find that reason documented anywhere. So: LGTM! :shipit:

PR #243, opened 2020-11-13
Pretty Print function for Transaction

### LINKED ISSUE #240: Add pretty-print function to Transaction
## Summary
A completed `Transaction` in the js-sdk is not the most human-readable format possible - for example `to` and `from` will display as byte arrays. Write a function on class `Transaction` that displays the object in a developer-convenient way, for example displaying `to` and `from` as string addresses.

## Scope
Does not break backwards compatibility, it is a new function on class `Transaction` in `transaction.js`.

## Urgency
Medium urgency: nothing is broken, but having this would enable dev-users to debug and develop much more quickly.

### See Also
https://github.com/algorand/js-algorand-sdk/issues/239

EvanJRichard:
## Summary 
If you have a built `Transaction`, then calling `console.log(theTxn)` does not give very useful information. Addresses are formatted for bytes, for example. This PR proposes to add a `prettyPrint()` function on class `Transaction` for user-dev debug.

I am especially looking for feedback on what fields should be edited and how. Currently the focus is on returning addresses to string, the genesis hash displays as b64, and the tag is returned to normal.

## Testing
Added a minimal unit test that checks if pretty print throws.

### See Also
Closes #240

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -591,6 +591,27 @@ class Transaction {
             this.fee += (ALGORAND_TRANSACTION_REKEY_LABEL_LENGTH + ALGORAND_TRANSACTION_ADDRESS_LENGTH) * feePerByte;
         }
     }
+
+    // pretty print the transaction to console
+    prettyPrint() {
+        let forPrinting = {
+            ...this
+        };
+        forPrinting.tag = forPrinting.tag.toString();
+        forPrinting.from = address.encodeAddress(forPrinting.from.publicKey);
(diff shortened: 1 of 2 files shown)


jasonpaulos:
Have you thought of overriding `toString()` instead of/in addition to this new method? Developers would already be familiar with that.

EvanJRichard:
@jasonpaulos good idea - I'll add a `toString()` that builds the same sort of dict, but instead of printing it, calls `JSON.stringify` and returns that

jasonpaulos (approved):
Looks good!

PR #246, opened 2020-11-19
MakeFoo family of functions get alternate FromObject method of calling

### LINKED ISSUE #159: Discussion/Enhancement: Allow object arguments as inputs to asset functions
This is a quality of life feature that I wanted to get some feedback on. 

The setup:
We've been working on a web application that allows users to configure and kick off asset transactions via a web-based javascript single-page application. For performance and security reasons, we're not including the `js-algorand-sdk` in the browser. We instead run it on a separate node service that's only accessible via secure APIs. Like most web apps these days, we're POSTing to REST-based endpoints with json payloads to provide our asset configuration data, and then our node service is using the `js-algorand-sdk` to generate additional fields and sign transactions. By design, our browser application doesn't interact with nodes directly. 

The "problem": 
I have to unpack the configuration object that's sent over the wire and break out each value into the multiple arguments that each asset SDK function takes in (For example, the createAsset action takes a total of 18 parameters where order is essential). These arguments are converted back into a standard js transaction object by the SDK functions anyways so it seems to be an unncessary step. There's also the old [Uncle Bob argument](https://www.matheus.ro/2018/01/29/clean-code-avoid-many-arguments-functions/) about keeping arguments to a minimum.

My suggestion: 
A standard object-based payload is arguably cleaner and streamlines translation issues between two services that a majority of the time (in javascript at least) relies on objects for inter-service transfer. I wrote some simple transaction schema validators in [algo-validation-agent](https://github.com/algonaut/algo-validation-agent) to help prevent errors that the chain would reject at the client side.

Would love to hear your thoughts about whether or not this would be a good change.

EvanJRichard:
Closes #159


EvanJRichard:
@jasonpaulos do you know anything about this error? https://travis-ci.com/github/algorand/js-algorand-sdk/jobs/443837553

jasonpaulos:
Yes, that's because a new chrome version came out that broke compatibility. I have a fix for this in #247.

EvanJRichard:
@jasonpaulos 🤔  the build is still failing, maybe I need to refresh something somewhere

jasonpaulos:
@EvanJRichard if you rebase off develop, it should be fixed.

EvanJRichard:
@jasonpaulos nice, I thought restarting the build would be enough alone. thanks!

jasonpaulos (approved):
Looks great! Thanks for taking the time to write out all those args

PR #247, opened 2020-11-19
Switch to webpack

### LINKED ISSUE #244: Webpack build error in algosdk 1.8.0
Using yarn berry to build a React application with an algosdk dependency causes the following error.

```sh
Δ yarn set version from sources
Δ yarn 
Δ yarn build
.yarn/cache/algosdk-npm-1.8.0-0f9313ab3a-3db2b154e4.zip/node_modules/algosdk/src/logicsig.js
Module not found: Recursion in resolving
```

Currently the fix is the prevent the version 1.8.0 from being installed.
```json
    "algosdk": ">=1.6.2 <1.8.0",
```

jasonpaulos:
Use webpack instead of browserify to build our browser bundle. Advantages of this are:

* More control over dependency versions. Browserify ships with a suite of node packages ported to the browser, but we have no control over specific versions. For instance, their assert module is stuck on an older version and broken for typed arrays.
* Explicitly declaring dependencies. Now users who build their own browser bundles can use our library without creating a custom config. This includes React developers.
* Better support for source maps. Source maps are much easier to produce to webpack, and I have enabled them to be generated.

There are some additional changes in this PR:
* I switched dependencies from `keccak` to `js-sha3` since `keccak` has problems bundling with webpack.
* I removed the use of the `assert` package from our source code, since it crashes if `process.env` isn't defined, which is no longer the case without browserify.
* Updated chromedriver, since a new stable version of chrome has been released.

Closes #244.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 
 .idea/
 .vscode
@@ -11,4 +12,4 @@ temp
 
 tests/cucumber/features/
 tests/cucumber/browser/build
-tests/browser/bundle.js
+tests/browser/bundle.*
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,761 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@babel/code-frame": {
-      "version": "7.5.5",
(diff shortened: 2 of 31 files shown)


EvanJRichard (approved):
Is this ready for review? Because it already LGTM :shipit:

jasonpaulos:
Yep, it's ready now

PR #253, opened 2020-12-01
Validate that assetMetaDataHash is 32 bytes

### LINKED ISSUE #250: Submitting an empty "assetMetadataHash" causes the signature validation to fail
For instance the following code snippet:
```typescript
const txnParams = {
      'from': addr,
      'note': note,
      'suggestedParams': suggestedParams,
      'assetTotal': supply,
      'assetDecimals': decimals,
      'assetDefaultFrozen': false,
      'assetUnitName': unitName,
      'assetName': assetName,
      'assetURL': assetURL,
      'assetMetadataHash': '',
      'assetManager': managerAddress,
      'assetReserve': reserveAddress,
      'assetFreeze': freezeAddress,
      'assetClawback': clawbackAddress,
      'type': 'acfg' as const,
    }
    const signedTxn = sdk.signTransaction(txnParams, sk)
    const transaction = await algodv2.sendRawTransaction(signedTxn.blob).do()
```

Will fail with:
```
{"message":"signature validation failed"}
```

If you remove the `assetMetadataHash` then this will then succeed happily.

jasonpaulos:
Currently, the Javascript SDK does not validate `assetMetadataHash` properly. This value is supposed to be exactly 32 bytes long, but if you pass in a smaller value the Javascript SDK will construct and sign a transaction with that smaller value.

Then when the transaction reaches the server, the `assetMetadataHash` gets unpacked into [the field here](https://github.com/algorand/go-algorand/blob/1424855ad2b5f6755ff3feba7e419ee06f2493da/data/basics/userBalance.go#L354), which is 32 bytes long. In the case of a smaller value, it appears extra 0s are added. This means that when the server tries to verify the transaction signature it fails because a different transaction with a shorter `assetMetadataHash` was actually signed in Javascript.

This PR fixes this by throwing an error if `assetMetadataHash` isn't 32 bytes long, and I've made it so that an empty string or a zero-length Uint8Array also get treated the same as `undefined` for this field.

Closes #250.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -94,6 +94,18 @@ class Transaction {
                 if (!Number.isSafeInteger(foreignAssetIndex) || foreignAssetIndex < 0) throw Error("each foreign asset index must be a positive number and smaller than 2^53-1");
             });
         }
+        if (assetMetadataHash !== undefined && assetMetadataHash.length !== 0) {
+            if (typeof(assetMetadataHash) === 'string') {
+                const encoded = Buffer.from(assetMetadataHash);
+                if (encoded.byteLength !== 32) {
+                    throw Error("assetMetadataHash must be a 32 byte Uint8Array or string.");
+                }
+                assetMetadataHash = new Uint8Array(encoded);
+            } else if (assetMetadataHash.constructor !== Uint8Array || assetMetadataHash.byteLength !== 32)
+                throw Error("assetMetadataHash must be a 32 byte Uint8Array or string.");
+        } else {
+            assetMetadataHash = undefined;
+        }
         if (note !== undefined) {
             if (note.constructor !== Uint8Array) throw Error("note must be a Uint8Array.");
         }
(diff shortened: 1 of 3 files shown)


src/transaction.js
```diff
@@ -94,6 +94,18 @@ class Transaction {
...
+                const encoded = Buffer.from(assetMetadataHash);
+                if (encoded.byteLength !== 32) {
+                    throw Error("assetMetadataHash must be a 32 byte Uint8Array or string.");
+                }
+                assetMetadataHash = new Uint8Array(encoded);
+            } else if (assetMetadataHash.constructor !== Uint8Array || assetMetadataHash.byteLength !== 32)
+                throw Error("assetMetadataHash must be a 32 byte Uint8Array or string.");
+        } else {
+            assetMetadataHash = undefined;
+        }
```
algorotem, line 108: The comment of the original function specifies that it needs to be a string. We should add a comment that specifies that it must be 32 bytes long. 
Also, since we specify it's a string, consider accepting only that (I guess with strict b64 form?).

algorotem (changes requested):

algorotem (approved):

PR #260, opened 2020-12-11
Support for decoding BigInts

### LINKED ISSUE #251: Large app state integers are rounded
## Summary
When an app has a number larger than Javascript's max safe integer, such as 79999999999999973, stored as a value in app state, the goal app read will output:

```
      "reserve": {
        "tt": 2,
        "ui": 79999999999999973
}
```

But calling `javascriptAlgosdkClient.accountInformation(account.addr).do()` to get the same app's state will return a rounded value of 79999999999999970.

## Scope
We'll need to put together a reproduction of this issue to see where the problem is. I suspect it's because `JSON.parse` will round any values greater than 2^53-1. If that's the case, the requests we send for any endpoint that returns app state should be encoded in msgpack, since that encoding can handle 64 bit ints.

## Urgency
This affects smart contracts that are already deployed. Should fix ASAP.

jasonpaulos:
Several of the aglod REST APIs can return JSON responses with integer values greater than JavaScript's maximum safe integer value. To fix this problem, this PR uses the [json-bigint](https://www.npmjs.com/package/json-bigint) library to parse the JSON response from our endpoints. This PR introduces the following modes to parse integers:
* `"default"`: Integers will be decoded according to `JSON.parse`, meaning they will all be Numbers and any values greater than `Number.MAX_SAFE_INTEGER` will lose precision.
* `"safe"`: All integers will be decoded as Numbers, but if any values are greater than `Number.MAX_SAFE_INTEGER` an error will be thrown.
* `"mixed"`: Integers will be decoded as Numbers if they are less than or equal to `Number.MAX_SAFE_INTEGER`, otherwise they will be decoded as BigInts.
* `"bigint"`: All integers will be decoded as BigInts.

To specify which mode you want to use, call the `setIntDecoding` method on a JSON request object. For example:
```javascript
const accountInfo = await client.accountInformation(account.addr).setIntDecoding('mixed').do();
```

If `setIntDecoding` is not called for a request, by default the `"default"` mode will be used. To make this feature easy to adopt and to reduce repetitiveness, the default mode for a client can be changed by calling the `setIntDecoding` method on an Algodv2 or Indexer client. For example:

```javascript
const client = new algosdk.Algodv2(...);
client.setIntDecoding('mixed'); // now all requests from this client will use the 'mixed' mode unless overridden

// this will do the same thing as the previous example
const accountInfo = await client.accountInformation(account.addr).do();
```

Fixes #251.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -573,6 +573,11 @@
       "integrity": "sha1-Qpzuu/pffpNueNc/vcfacWKyDiA=",
       "dev": true
     },
+    "bignumber.js": {
+      "version": "9.0.1",
+      "resolved": "https://registry.npmjs.org/bignumber.js/-/bignumber.js-9.0.1.tgz",
+      "integrity": "sha512-IdZR9mh6ahOBv/hYGiXyVuyCetmGJhtYkqLBpTStdhEGjegpPlUawydyaF3pbIOFynJTpllEs+NP+CS9jKFLjA=="
+    },
     "bluebird": {
       "version": "3.7.2",
       "resolved": "https://registry.npmjs.org/bluebird/-/bluebird-3.7.2.tgz",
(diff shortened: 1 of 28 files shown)


robert-zaremba:
When this will get merged? We need this in https://github.com/scale-it/algorand-builder

jasonpaulos:
@robert-zaremba I have some concerns that this PR introduces breaking changes to the current behavior and some inconsistencies between how BigInts are returned in JSON responses vs msgpack responses. I plan on revisiting this PR next week to address those concerns. For the time being I'm going to update this to a draft PR.

jasonpaulos:
We've decided on a new approach for this fix that maintains compatibility with existing user code:
* Make no change to existing requests such as `client.accountInformation(account.addr).do()`.
* Add a new option in the request builder to specify how integers should be decoded, for example `client.accountInformation(account.addr).integerDecoding('safe').do()`. Some of the supported options will be:
    * `safe`: decode everything as numbers and throws an error if there's an integer larger than Number.MAX_SAFE_INTEGER.
    * `bigint`: decode everything as BigInts.
    * `smart/mixed`: decode integers larger than Number.MAX_SAFE_INTEGER as BigInts and everything else as numbers.

It could also make sense to add the ability to change the default `integerDecoding` value for all requests on a given `Algodv2` client so users would only need to specify their choice in one place.

The above is all in the context of JSON responses, but I'm going to think about how we can make this behavior work for msgpack encoded responses too.

jasonpaulos:
@rotemh this is ready to review again.

algorotem (approved):
Approved, minus the excluding the discussion we had.

PR #267, opened 2021-01-15
Fixing rekeyTo makeTxnWithSuggestedParams bug

Alex99y:
The current implementation of these functions initializes the object with the field rekeyTo  (lowercase letter K) causing the build of the rekey transaction to fail because the expected field name is reKeyTo (uppercase letter K).

- makePaymentTxnWithSuggestedParams
- makeKeyRegistrationTxnWithSuggestedParams
- makeAssetCreateTxnWithSuggestedParams
- makeAssetConfigTxnWithSuggestedParams
- makeAssetDestroyTxnWithSuggestedParams
- makeAssetFreezeTxnWithSuggestedParams
- makeAssetTransferTxnWithSuggestedParams

To avoid this problem in the future, reKeyTo has been changed by rekeyTo in src/makeTxn.js and src/transaction.js (As it is in the node go-algorand)

### DIFF
--- a/src/makeTxn.js
+++ b/src/makeTxn.js
@@ -56,7 +56,7 @@ function makePaymentTxnWithSuggestedParams(from, to, amount, closeRemainderTo, n
         "note": note,
         "suggestedParams": suggestedParams,
         "type": "pay",
-        "rekeyTo": rekeyTo
+        "reKeyTo": rekeyTo
     };
     return new txnBuilder.Transaction(o);
 }
@@ -134,7 +134,7 @@ function makeKeyRegistrationTxnWithSuggestedParams(from, note, voteKey, selectio
         "voteKeyDilution": voteKeyDilution,
         "suggestedParams": suggestedParams,
         "type": "keyreg",
-        "rekeyTo": rekeyTo
+        "reKeyTo": rekeyTo
     };
     return new txnBuilder.Transaction(o);
 }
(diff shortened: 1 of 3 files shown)


jasonpaulos (changes requested):
Thanks for noticing this inconsistency! I agree that `rekey` makes more sense than `reKey` does currently, but at the time rekeying was introduced to this package there wasn't yet a consensus on this. Unfortunately to avoid introducing a breaking change I think we need to make sure the `reKeyTo` property still works in the `Transaction` class. In v2 of this library we'd definitely want to use `rekey` though.

So I think there are two options here:
1. Make all makeTxn calls pass `reKey` to the Transaction constructors and leave Transaction alone.
2. Make all makeTxn calls pass `rekey` to the Transaction constructor, modify the Transaction class to use `rekey` internally, and put in a setter and getter for `reKey` (the old spelling) to avoid breaking existing user code.

If you have another idea of how to avoid a breaking change, let me know. Otherwise one of the above changes would have to be made before I can approve this.

Alex99y:
Perfect. I chose the first option, we make it less confusing. I agree to use the rekey for the next version.

Alex99y:
@jasonpaulos some tests are using _rekeyTo_, tomorrow I can take a look at all failed tests.

jasonpaulos (approved):
Everything looks good, thanks @Alex99y!

PR #269, opened 2021-01-21
decodeAddress now verifies the checksum (fixes #262)

### LINKED ISSUE #262: decodeAddress should verify the checksum
# Summary

The `decodeAddress(address)` function parses the public key and checksum but does not verify that the checksum is correct. The function already throws an exception if the address is invalid, so we should throw an exception if the checksum is invalid as well.

Neurone:
Fixed also an npm command in readme

Fixes #262

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Include this line in your HTML.
 
 To build a new version of the library for browsers, run:
 ```bash
-npm build
+npm run build
 ```
 
 ### Testing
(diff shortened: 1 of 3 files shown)


jasonpaulos (changes requested):
Thanks for implementing this @Neurone, this looks great! Could you add a test in `tests/3.Address.js` to verify `decodeAddress` throws an error when given an address with an invalid checksum (probably with `assert.throws(...)`)? I'd be happy to accept once that's added.

Neurone:
Hi @jasonpaulos, sure. I'll send an update during this weekend.

jasonpaulos (approved):
Looks great! Thanks for working on this.

PR #277, opened 2021-02-04
Enable Custom HTTPClients for Algod and Indexer clients

janmarcano:
Allows for a custom HTTPClient to be provided for Algod and Indexer clients during creation


jasonpaulos:
Hi @janmarcano, thanks for contributing. I'm curious, what is the added benefit or use case for passing in a custom HTTP client?

janmarcano:
Hi @jasonpaulos !

The purpose would be to enable the possibility for external Algorand apps/client to integrate themselves into the SDK calls.

An example would be to use AlgoSigner to handle the API calls:
```
const c = AlgoSigner.getAlgodHTTPClient('TestNet');
const algod = new algosdk.Algodv2(null, null, null, null, c);
await algod.getTransactionParams().do();
```

This way, all calls are made using AlgoSigner's backend.

algorotem:
Hi @janmarcano,

Thanks for the PR. We believe that enabling the custom client without a proper layer of abstraction will bind us to a specific set and may make future handling difficult and therefor I can't approve this PR. 
We suggest wrapping the the creation of the client with another function that also overrides the internal one. This will also provide a cleaner API for the users. 
e.g. 
```javascript
AlgoSigner.makeAlgodClient(...) 
```
where internally you have 
```javascript
algod.c = CustomClient
```

PR #280, opened 2021-02-19
Fix encoding of all 0 lease and assetMetadataHash

### LINKED ISSUE #272: Make sure the SDK and algod agree on encoding
## Summary
If the `lease` or `assetMetadataHash` fields of `Transaction` are encoded in a way inconsistant with algod, signatures generated by this SDK will appear to be invalid to algod.

## Scope
Make sure that the SDK correctly validates and omits (when empty) the `lease` and `assetMetadataHash` fields.

## Urgency
This is a small but important fix.

jasonpaulos:
This PR fixes an encoding inconsistency between this SDK and algod when `lease` or `assetMetadataHash` consist of all zeros. A similar fix for the Python SDK was recently merged: algorand/py-algorand-sdk#174.

Fixes #272.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -371,20 +371,24 @@ export class Transaction implements TransactionStorageStructure {
       txn.assetMetadataHash.length !== 0
     ) {
       if (typeof txn.assetMetadataHash === 'string') {
-        const encoded = Buffer.from(txn.assetMetadataHash);
-        if (encoded.byteLength !== ASSET_METADATA_HASH_LENGTH) {
-          throw Error(
-            `assetMetadataHash must be a ${ASSET_METADATA_HASH_LENGTH} byte Uint8Array or string.`
-          );
-        }
-        txn.assetMetadataHash = new Uint8Array(encoded);
-      } else if (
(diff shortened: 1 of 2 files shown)


src/transaction.js
```diff
@@ -101,13 +101,17 @@ class Transaction {
...
                 throw Error("assetMetadataHash must be a " + ASSET_METADATA_HASH_LENGTH + " byte Uint8Array or string.");
+            }
+
+            if (assetMetadataHash.every(value => value === 0)) {
+                // if hash contains all 0s, omit it
+                assetMetadataHash = undefined;
+            }
         } else {
             assetMetadataHash = undefined;
         }
```
algorotem, line 117: Since the code conditioning on the value of `assetMetadaataHash`, consider having as a sequence of if statements and avoid the `else`s

algorotem (changes requested):

src/transaction.js
jasonpaulos: I think this section of code already has the minimum number of `else`s.

jdtzmn (approved):
Looks good!

PR #283, opened 2021-03-01
don't send X-Algo-API-Token header when token is not set

adkri:
Currently sends an empty  X-Algo-API-Token header even with the client is initialized with token set to null.

Not all services allow for this header to be set, so don't include this header when the token is null.


jasonpaulos (changes requested):
Thanks for working on this, it looks like a nice improvement!

One small thing: could you remove the changes to dist folder? We normally only update those once a release is ready.

jasonpaulos:
Also I believe @amitadkri will need to sign the CLA before I can accept this.

PR #289, opened 2021-03-04
Update examples to V2

### LINKED ISSUE #254: Examples needs to be updated to V2
## Summary
Most of the examples in this repo are outdated and need to be updated to V2. This includes the documentation in the readme and the sample web app.

## Scope
We should update the entire set of examples similarly to what has already happened to the python sdk: https://github.com/algorand/py-algorand-sdk/tree/develop/examples.

Some sections that the examples should cover:
* Using the transaction note field
* Creating/sending group transactions
* Using multisigs
* Creating LogicSigs and signing transactions with them
* Specific asset transactions (create, freeze, clawback, etc.)
* Specific app transactions (create, opt in, call, etc.)

## Urgency/Relative priority
We're showing incorrect information on the documentation, it's something we should change asap. Could result in bad states for users.

jdtzmn:
This PR fixes issues with the existing web app and adds a standard set of example scripts to the `examples` directory.

Fixes #254.


examples/app_transaction_examples.js
```diff
@@ -0,0 +1,141 @@
...
+  const callTxn = algosdk.makeApplicationNoOpTxn(receiver.addr, suggestedParams, appId, appArgs);
+
+  // send the transaction
+  logBold('Sending application call transaction.');
+  const signedCallTxn = callTxn.signTxn(receiver.sk);
+  const { txId: callTxnId } = await client.sendRawTransaction(signedCallTxn).do();
+
+  // wait for confirmation
+  await verboseWaitForConfirmation(client, callTxnId);
+
```
jasonpaulos, line 123: Can you add an example of a close out call as well?

examples/asset_create_example.js
```diff
@@ -0,0 +1,65 @@
...
+  const feePerByte = 10;
+  const firstValidRound = 1000;
+  const lastValidRound = 2000;
+  const genesisHash = 'SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=';
+
+  const total = 100; // how many of this asset there will be
+  const assetName = "assetname";
+  const unitName = "unitname";
+  const url = "website";
+  const metadata = "fACPO4nRgO55j1ndAK3W6Sgc4APkcyFh"; // should be a 32-byte hash
```
jasonpaulos, line 26:
```suggestion
  const metadata = new Uint8Array(Buffer.from("664143504f346e52674f35356a316e64414b3357365367633441506b63794668", "hex")); // should be a 32-byte hash
```
  
Something like this would be a better example, since we really want to encourage passing in a Uint8Array instead of a string for this parameter.
 
Also, can you include `decimals` as a parameter in this section too?

examples/log_sig_example.js
```diff
@@ -0,0 +1,53 @@
+// Example: creating a LogicSig transaction signed by a program that never approves the transfer.
```
jasonpaulos, line 1: I think this file should be `logic_sig_example.js`?

examples/app_transaction_examples.js
```diff
@@ -0,0 +1,141 @@
...
+  const numLocalByteSlices = 0;
+  const numGlobalInts = 0;
+  const numGlobalByteSlices = 0;
+  const appArgs = [];
+
+  // get suggested params
+  const suggestedParams = await client.getTransactionParams().do();
+
+  // create the application creation transaction
+  const createTxn = algosdk.makeApplicationCreateTxn(
```
jasonpaulos, line 72: Nit: could you use the `make___TxnFromObject` methods in this example too?

jasonpaulos (changes requested):
This is a much-needed improvement! I've left a few small comments, but besides those everything looks good.

jasonpaulos (approved):

PR #291, opened 2021-03-05
Initialize errors when they occur for improved error traces

### LINKED ISSUE #286: Initialize errors when needed for a more useful trace
Example code:

https://github.com/algorand/js-algorand-sdk/blob/0640240ba167a52c3f05a8d7df1e425df4f9c616/src/main.js#L31

There are a several locations like above where errors are initialized before they are needed. By constructing the error this way, the trace does not reference where the error occurred, it only display information regarding where the error was constructed:

```bash
    at Object.<anonymous> (/Documents/GitHub/js-algorand-sdk/src/encoding/address.js:13:33)
    at Module._compile (node:internal/modules/cjs/loader:1091:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1120:10)
    at Module.load (node:internal/modules/cjs/loader:971:32)
    at Function.Module._load (node:internal/modules/cjs/loader:812:14)
    at Module.require (node:internal/modules/cjs/loader:995:19)
    at require (node:internal/modules/cjs/helpers:92:18)
    at Object.<anonymous> (/Documents/GitHub/js-algorand-sdk/src/main.js:3:17)
    at Module._compile (node:internal/modules/cjs/loader:1091:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1120:10)
```

These constructed errors are often exported, and so a solution must preserve that functionality as well. A possible fix to this issue would be to export "error builders" like so:

```js
function constructMultisigBadSenderError() {
	return new Error("The transaction sender address and multisig preimage do not match.")
}
```

The trace is much more helpful as a result:

```bash
    at decodeAddress (/Documents/GitHub/js-algorand-sdk/src/encoding/address.js:42:15)
    at /Documents/GitHub/js-algorand-sdk/src/encoding/address.js:121:16
    at Array.map (<anonymous>)
    at Object.fromMultisigPreImgAddrs (/Documents/GitHub/js-algorand-sdk/src/encoding/address.js:120:23)
    at Object.multisigAddress (/Documents/GitHub/js-algorand-sdk/src/main.js:185:20)
    at main (/Documents/GitHub/js-algorand-sdk/examples/multisig_example.js:29:31)
    at Object.<anonymous> (/Documents/GitHub/js-algorand-sdk/examples/multisig_example.js:52:1)
    # ^-- Notice that this line includes the position in client code where the error occurred
    at Module._compile (node:internal/modules/cjs/loader:1091:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1120:10)
    at Module.load (node:internal/modules/cjs/loader:971:32)
```

jdtzmn:
Fixes #286 by setting error *messages* instead of initialized Error instances. This allows for error traces to be more informative to the user. Please refer to #286 for additional details.

### DIFF
--- a/src/convert.js
+++ b/src/convert.js
@@ -1,5 +1,5 @@
 const MICROALGOS_TO_ALGOS_RATIO = 1e6;
-const ERROR_INVALID_MICROALGOS = new Error("Microalgos should be positive and less than 2^53 - 1.");
+const INVALID_MICROALGOS_ERROR_MSG = "Microalgos should be positive and less than 2^53 - 1.";
 
 /**
  * microalgosToAlgos converts microalgos to algos
@@ -8,7 +8,7 @@ const ERROR_INVALID_MICROALGOS = new Error("Microalgos should be positive and le
  */
 function microalgosToAlgos(microalgos) {
     if (microalgos < 0 || !Number.isSafeInteger(microalgos)){
-        throw ERROR_INVALID_MICROALGOS;
+        throw new Error(INVALID_MICROALGOS_ERROR_MSG);
     }
     return microalgos/MICROALGOS_TO_ALGOS_RATIO
 }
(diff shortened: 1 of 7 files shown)


jasonpaulos (approved):
Looks good!

PR #293, opened 2021-03-09
Update multisig code

### LINKED ISSUE #292: Update multisig code
## Summary
The [SDK multisig code](https://github.com/algorand/js-algorand-sdk/blob/0640240ba167a52c3f05a8d7df1e425df4f9c616/src/multisig.js#L22-L70) has not been updated in a long time, and it currently does not support non-payment transactions very well. We should update this code and add new tests to verify multisig works as intended for non-payment transactions.

## Scope
* Remove overrides of `MultisigTransaction`'s `get_obj_for_encoding` and `from_obj_for_encoding` that treat non-payment transactions differently.
* Verify read-only methods (e.g. `estimageSize`, `txID`, `toString`) on `Transaction` work for `MultisigTransaction`s of any type.
* Override methods that mutate the transaction (e.g. `addLease`, `addRekey`) on `MultisigTransaction` to throw an error.
* Override `signTxn` on `MultisigTransaction` to throw an error.
* Add tests to verify multisig signing works properly on non-payment transactions.

## Urgency
There is currently no SDK testing for a non-payment multisig transaction, so it's important to do this.

jdtzmn:
This PR removes `MultisigTransaction`'s unnecessary overrides of `get_obj_for_encoding` and `from_obj_for_encoding`, verifies that read-only methods work for multisig transactions of any type, and overrides certain inherited methods that are prohibited in the context of multisig transactions.

Closes #292.


jasonpaulos (changes requested):
Looks good except for one small thing!

Can you remove the `if (!this.hasOwnProperty("objForEncoding"))` statement from `partialSignTxn`? We should keep the body but the conditional is always true now.

jdtzmn:
Good catch!

jasonpaulos (approved):
Looks great! The build errors are not because of these changes, so I'm merging anyway.

PR #325, opened 2021-03-25
Generate API types using code generation

### LINKED ISSUE #311: Use code generation to create API types
## Summary
Similar to [go-algorand-sdk](https://github.com/algorand/go-algorand-sdk) and [java-algorand-sdk](https://github.com/algorand/java-algorand-sdk), this SDK can use code generation to create request and response types for the algod and indexer APIs.

## Scope
Create and run code generation templates to generate TypeScript files for algod and indexer responses and requests.

## Urgency
Required to finish the TypeScript migration.

jdtzmn:
This PR includes a code template for API types and generated response types for the algod and indexer APIs.

Closes #311.


src/types/blockHeader.d.ts
```diff
@@ -0,0 +1,71 @@
+/**
```
jasonpaulos, line 1: I think this file ending should be `.ts`.

src/client/v2/algod/models/types.ts
```diff
@@ -0,0 +1,1390 @@
...
+    protocolVersion,
+    round,
+    sources,
+    txns,
+  }: {
+    accounts: Account[];
+    apps: Application[];
+    latestTimestamp: number;
+    protocolVersion: string;
+    round: bigint;
```
jasonpaulos, line 672: I wonder if we should make the `uint64` and `integer` API types map to `number | bigint`? They should both get encoded the same way by the SDK, so I think this change would make the types more flexible to "input" types like this that users create.

jasonpaulos (commented):
Mostly looks good!

I noticed that the new generated types aren't used yet, but I think we can make a separate issue for that.

src/client/v2/algod/models/types.ts
jdtzmn: Yes, I think that makes sense.

jasonpaulos (approved):

PR #326, opened 2021-03-26
Remove `dist` directory

### LINKED ISSUE #324: Delete the `dist` folder
## Summary
The `dist` folder is getting pretty large, especially since the output of the TypeScript compilation process is being put there. The original reason for including this folder was to provide an easy way for users to download minified browser bundles. We should stop checking in this folder to the repo and address that issue in a better way.

## Scope
The proposal is:
* Delete the `dist` folder from the repo and put it in the `.gitignore` (and maybe also `.dockerignore` 🤔).
* For each GitHub release we have, retroactively generate bundles (or steal from previous versions in the repo history) and upload them as release assets.
* Update our readme to direct users toward the GH release page to download minified browser bundles.
* Create a new file to document the publishing process. This should include:
  * The commands needed to fully populate the `dist` folder before publishing (i.e. how to generate bundles and types from source).
  * ~Directions about which files need to be uploaded to new GH releases.~ Future releases should use CDNs, so maybe we can include a link to a CDN in the GH release?

## Urgency
This blocks #323.

jdtzmn:
This PR removes the `dist` directory from version control and includes instructions for finding previous assets.

Some tasks that will need to be done in the future:

- [ ] Add documentation for the publishing process. This will largely depend on how releases are published in the future.
- [ ] Include a link to a CDN in the README.md right before new releases have been published.

Closes #324.


jasonpaulos (approved):
Looks great

PR #332, opened 2021-03-31
Use node’s `url` module for URL construction in the HTTPClient constructor

### LINKED ISSUE #261: Port in algod/indexer does not work with URL starting with some path
The port is concatenated to the baseserver URL on https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/client.js#L41.

This means that it does not work properly when baseserver contains some path beyond just the server.

Concretely, to use algoexplorer.io indexer, you need to write
```js
const algodIndexer = new algosdk.Indexer("", "https://api.testnet.algoexplorer.io/idx2", "");
```
but not:
```js
const algodIndexer = new algosdk.Indexer("", "https://api.testnet.algoexplorer.io/idx2", 443);
```
nor:
```js
const algodIndexer = new algosdk.Indexer("", "https://api.testnet.algoexplorer.io/idx2");
```

Maybe the best solution is to add a comment in the documentation of the constructor of the indexer and algod clients.

jdtzmn:
Fixes #261.

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -1,3 +1,4 @@
+import { URL } from 'url';
 import request from 'superagent';
 import * as utils from '../utils/utils';
 import IntDecoding from '../types/intDecoding';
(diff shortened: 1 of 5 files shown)


src/client/client.ts
```diff
@@ -120,16 +122,28 @@ export default class HTTPClient {
...
+    const address = new Url(
+      path.posix.join(this.baseURL.pathname, relativePath),
+      this.baseURL
+    );
+    return address.toString();
+  }
+
   /**
    * Send a GET request.
    * @param {string} path The path of the request.
```
jasonpaulos, line 149:
```suggestion
   * @param {string} relativePath The path of the request.
```

jasonpaulos (approved):
Jus one small docs nit, feel free to merge after fixing.

PR #337, opened 2021-04-07
Final TypeScript changes

### LINKED ISSUE #336: Final TypeScript Changes
## Summary
This issue tracks a few final TypeScript changes that require fixing before the TypeScript beta can be released.

## Scope
- [x] `assignGroupID` method requires a better return type
- [x] Incorrect default generic types for the `JSONRequest` abstract class 
- [x] `Account` type should be exported
- [x] Specify `SuggestedParams` type for the suggested parameters client request
- [x] Improve `makeKeyRegistrationTxn(...)` type definitions using overloads
- [x] Warn if trying to create a client with a base server of `localhost`, since `localhost` is not considered a valid URL

## Urgency
Requires for TypeScript beta release.

jdtzmn:
* Remove deprecated ESLint setting

* Specify `SuggestedParams` type for the SuggestedParams client request

* Fix default generic types of the `JSONRequest` class

* Provide better return type for the `assignGroupID` method

* Export `Account` type

Closes #336.

### DIFF
--- a/src/makeTxn.ts
+++ b/src/makeTxn.ts
@@ -336,7 +336,7 @@ export function makeKeyRegistrationTxnWithSuggestedParamsFromObject(
     >,
     'from' | 'note' | 'suggestedParams' | 'rekeyTo'
   > & {
-    nonParticipation?: true;
+    nonParticipation: true;
   }
 ): txnBuilder.Transaction;
 export function makeKeyRegistrationTxnWithSuggestedParamsFromObject(o: any) {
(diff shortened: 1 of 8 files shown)


src/makeTxn.ts
```diff
@@ -262,9 +323,24 @@ export function makeKeyRegistrationTxnWithSuggestedParamsFromObject(
...
+export function makeKeyRegistrationTxnWithSuggestedParamsFromObject(
+  o: Pick<
+    RenameProperty<
+      MustHaveSuggestedParams<KeyRegistrationTxn>,
+      'reKeyTo',
+      'rekeyTo'
+    >,
+    'from' | 'note' | 'suggestedParams' | 'rekeyTo'
+  > & {
+    nonParticipation?: true;
```
jasonpaulos, line 339:
```suggestion
    nonParticipation: true;
```
I am seeing the following code pass incorrectly, but I believe this change should fix it:
```javascript
algosdk.makeKeyRegistrationTxnWithSuggestedParamsFromObject({
    from: account.addr,
    // none of the key params or nonParticipation
    suggestedParams: await client.getTransactionParams().do(),
})
```

jasonpaulos (approved):
Pending tests, looks great!

PR #347, opened 2021-04-21
Allow byte arrays in algod compile

jasonpaulos:
This small change allows `Uint8Arrays` to be passed into the algod `compile` method.


jdtzmn (approved):
Looks good!

PR #359, opened 2021-05-03
Fix asset creation transaction types that should be optional

### LINKED ISSUE #358: Type Error in Asset Creation method
Hi,

there seems to be a type error for the `makeAssetCreateTxnWithSuggestedParamsFromObject(params)` function parameter in TypeScript. It requires the `manager`, `freeze`, `clawback` and `reserve` fields. 

Due to the [docs](https://developer.algorand.org/docs/features/asa/) these fields are optional and can be set to `""` to disable the functionality. However, setting the parameters to an empty string leads to the following error:
```
```

As a workaround cast the parameter type to 'any' ->`makeAssetCreateTxnWithSuggestedParamsFromObject(<any> params)`

jdtzmn:
This PR fixes fields such as  `assetManager` and `assetClawback` in the `TransactionParams` type that are incorrectly marked as required.

Closes #358.

### DIFF
--- a/src/types/transactions/base.ts
+++ b/src/types/transactions/base.ts
@@ -226,36 +226,36 @@ export interface TransactionParams {
   /**
    * String representation of Algorand address in charge of reserve, freeze, clawback, destruction, etc.
    */
-  assetManager: string;
+  assetManager?: string;
 
   /**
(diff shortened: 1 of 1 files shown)


jasonpaulos (approved):
Looks good

PR #360, opened 2021-05-07
Add SRI for security and misc changes for browser

fabrice102:
* SRI (subresource integrity) ensures that the CDN cannot serve a
  malicious script. This is quite important for the Algorand SDK
  as the Algorand SDK can be used in quite sensitive websites.
  See https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
* Use the full URL for algosdk.min.js in the CDN so that the
  source map is properly found by the browser which facilitates
  debugging
* Close the `<script>` tag using `</script>` for better compatibility
* Specify explicitely the cross-origin to be `anonymous`

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -9,3 +9,13 @@ In prior versions, the browser bundles were made available in the `dist` directo
 ## Can I host the browser bundle myself?
 
 Yes! If you would instead prefer to host the package yourself, a minified browser bundle can be found in the `dist/browser/` folder of the published npm package starting with version 1.9.0.
+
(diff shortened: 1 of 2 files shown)


jasonpaulos:
I apologize for the delay in reviewing this. @fabrice102 if you can update the SRI hash to v1.11.1 of this library, it looks good to merge

fabrice102:
@jasonpaulos updated to v1.11.1

jasonpaulos (approved):
Thank you!

PR #368, opened 2021-06-07
update langspec

shiqizng:
Regenerate langspec.json to include latest TealV4 opcodes.

PR #369, opened 2021-06-07
"rent" extra pages for app create

AyAggarwal:
Followed discussion from https://github.com/algorand/py-algorand-sdk/pull/196, put extraPages last in parameter order. Updated tests to explicitly populate undefined parameters for readability


src/transaction.ts
```diff
@@ -253,6 +255,10 @@ export class Transaction implements TransactionStorageStructure {
       throw Error('firstRound must be a positive number');
     if (!Number.isSafeInteger(txn.lastRound) || txn.lastRound < 0)
       throw Error('lastRound must be a positive number');
+    if (txn.extraPages < 0 || txn.extraPages > 3)
+      throw Error(
+        'extra-pages must be an Integer between and including 0 to 3'
```
AyAggarwal, line 260: all numbers in javascript are floats, should I manually check if the value is within {0,1,2,3}? or is this okay

jdtzmn: Perhaps use `Number.isInteger()` method?

jasonpaulos: If you add a `Number.isInteger` check in addition to this, that should be sufficient.

src/makeTxn.ts
```diff
@@ -1138,6 +1141,7 @@ export function makeApplicationCreateTxnFromObject(
         appForeignApps: 'foreignApps';
         appForeignAssets: 'foreignAssets';
         reKeyTo: 'rekeyTo';
+        extraPages: 'extraPages';
```
jdtzmn, line 1144: The `RenameProperties` TypeScript generic type renames properties but _passes properties as well_, so this assignment is unnecessary. Adding `'extraPages'` to the `Pick` union is sufficient.

tests/cucumber/steps/steps.js
```diff
@@ -4284,8 +4292,10 @@ module.exports = function getSteps(options) {
         appAccounts,
         appForeignApps: foreignApps,
         appForeignAssets: foreignAssets,
-        type: 'appl',
-        suggestedParams: sp,
+        note: undefined,
+        lease: undefined,
+        reKeyTo: undefined,
```
jdtzmn, line 4297: Must these be explicitly set to `undefined`? Does the test fail when these properties are not set?

jdtzmn (approved):
`Number.isInteger` should be added and I had a few questions, but otherwise this looks good!

src/transaction.ts
AyAggarwal: Ok I'll use that LOL

AyAggarwal: before it was giving me an error but it works now that the test file is fixed

src/makeTxn.ts
AyAggarwal: gotcha

tests/cucumber/steps/steps.js
AyAggarwal: In this case it may work because its from an object, but removing these parameters gives issues in other tests so I rather keep it consistent

jdtzmn: I personally disagree - I think it's generally bad practice to set undefined properties on an object, but I'll wait to see what @jasonpaulos thinks.

src/transaction.ts
jdtzmn, line 260: nit: change this to `extraPages` because that is the name of the parameter passed in.

AyAggarwal: done

tests/cucumber/steps/steps.js
jasonpaulos: In this case I'd say the `undefined` fields should be omitted since they have nothing to do with this test.

AyAggarwal: cool thanks yall

onetechnical:
You may need to adjust the chrome driver, as you have this error:

```
SessionNotCreatedError: session not created: This version of ChromeDriver only supports Chrome version 89
Current browser version is 91.0.4472.77 with binary path /usr/bin/google-chrome
    at Object.throwDecodedError (/home/travis/build/[secure]/js-[secure]-sdk/node_modules/selenium-webdriver/lib/error.js:550:15)
    at parseHttpResponse (/home/travis/build/[secure]/js-[secure]-sdk/node_modules/selenium-webdriver/lib/http.js:565:13)
    at Executor.execute (/home/travis/build/[secure]/js-[secure]-sdk/node_modules/selenium-webdriver/lib/http.js:491:26)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  remoteStacktrace: '#0 0x55db65b662b9 <unknown>\n'
}
```

jasonpaulos:
I've made #370 to update chromedriver and a few other deps, which should fix the CI issue. Hopefully it can get merged tomorrow and this PR can rebase/merge with develop.

src/types/transactions/base.ts
```diff
@@ -366,4 +366,9 @@ export interface TransactionParams {
    * All new Algorand accounts are participating by default. This means they earn rewards.
    */
   nonParticipation?: boolean;
+
+  /**
+   * Int representing extra pages of memory to rent during an application create transaction.
```
jasonpaulos, line 371: Could you mention that a page is 1024 bytes, like in https://github.com/algorand/py-algorand-sdk/blob/3ca47273258ae99e168272f95b66416359a00c97/algosdk/future/transaction.py#L1161?

jannotti:
It's a little more complicated than that now.  Because we're also combining
the limit on Approval and Clear, these pages are 2k, but shared.

jasonpaulos:
@jannotti so what are the effects of requesting 1 extra page? An additional 1024 bytes to the approval program, or an additional 2048 bytes to the shared pool of space between the approval & clear state program?

jannotti:
> @jannotti so what are the effects of requesting 1 extra page? An additional 1024 bytes to the approval program, or an additional 2048 bytes to the shared pool of space between the approval & clear state programs?

The latter.  You get 2 extra k, to be split as you like. So with 1 extra page you could have a nearly 4k Approval, and a tiny Clear State.

src/transaction.ts
```diff
@@ -418,6 +422,7 @@ export class Transaction implements TransactionStorageStructure {
     if (txn.selectionKey !== undefined) {
       txn.selectionKey = Buffer.from(txn.selectionKey as string, 'base64');
     }
+
```
jdtzmn, line 425:
```suggestion
```
nit: unnecessary newline in an otherwise un-spaced section of conditionals.

src/transaction.ts
```diff
@@ -253,6 +255,8 @@ export class Transaction implements TransactionStorageStructure {
       throw Error('firstRound must be a positive number');
     if (!Number.isSafeInteger(txn.lastRound) || txn.lastRound < 0)
       throw Error('lastRound must be a positive number');
+    if (txn.extraPages < 0 || txn.extraPages > 3)
```
jdtzmn, line 258: What was the issue with `Number.isInteger`?

AyAggarwal: It somehow causes all the tests to fail with extra pages parameter to fail. Odd because they are most definitely integers

AyAggarwal: It's because I didn't check for undefined first 🤦‍♂️

jasonpaulos:
```suggestion
    if (txn.extraPages != null && (!Number.isInteger(txn.extraPages) || txn.extraPages < 0 || txn.extraPages > 3))
```

jasonpaulos (changes requested):
Good except for the missing `Number.isInteger` check. I will approve immediately once that's added.

jasonpaulos:
Apologies, it seems my suggested change did not satisfy prettier or eslint ☹️

AyAggarwal:
> Apologies, it seems my suggested change did not satisfy prettier or eslint ☹️

That's okay I force pushed my local version

src/transaction.ts
```diff
@@ -253,6 +256,13 @@ export class Transaction implements TransactionStorageStructure {
       throw Error('firstRound must be a positive number');
     if (!Number.isSafeInteger(txn.lastRound) || txn.lastRound < 0)
       throw Error('lastRound must be a positive number');
+    if (
+      txn.extraPages != null &&
```
jdtzmn, line 260: I'm not sure this null check is necessary

jdtzmn (changes requested):

src/transaction.ts
AyAggarwal: in my original commit it checks for undefined, which is necessary because extrapages is only defined in create tests. otherwise the error gets triggered because undefined is not an integer

jasonpaulos: Right, I suppose `Number.isInteger` would return false for null, rending the null check a little redundant.

AyAggarwal: its not redundant because we don't want it to return false for undefined. this test should only apply to cases where the parameter was passed in

jasonpaulos: Great point!

AyAggarwal:
BTW I have #368 merged in here right now

PR #370, opened 2021-06-07
Update `chromedriver` & other deps

jasonpaulos:
Update `chromedriver` and several other dependencies. This should fix Chrome tests failing on travis.


jasonpaulos:
This will be incorporated into #369

PR #379, opened 2021-06-16
Use DryrunRequest instead of DryrunSource in constructor

jordpw:
This change addresses Issue algorand#366, which points out that the constructor for the `Dryrun` class should be of type `DryrunRequest`, instead of `DryrunSource`.


jasonpaulos (approved):
Thanks for making this!

PR #382, opened 2021-06-21
Update examples on getting `suggestedParams`

aldur:
Show how to let the SDK fetch suggested parameters as an alternative to setting them manually.


jasonpaulos (approved):
Looks good, thanks for adding this

PR #383, opened 2021-06-21
Use JSON for dryrun request body

jasonpaulos:
This PR changes the serialization from msgpack to JSON for dryrun requests. The advantages of this are:
* No more `Error: The object contains empty or 0 values` messages if a DryrunRequest object contains an empty value.
* TEAL programs are properly encoded if they are passed in as base64-encoded compiled program bytes.


jasonpaulos:
Update: we probably don't want to merge this. Rather, we should make the existing dryrun classes work better with msgpack encoding.

jasonpaulos:
Closing in favor of #399 instead.

PR #384, opened 2021-06-21
Fee pooling SDK change

shiqizng:
This PR is to support creating transactions with a fee under mintxnfee since other txns in the group will be paying.

### DIFF
--- a/src/client/v2/algod/models/types.ts
+++ b/src/client/v2/algod/models/types.ts
@@ -1685,8 +1685,8 @@ export class TransactionParametersResponse extends BaseModel {
   /**
    * Fee is the suggested transaction fee
    * Fee is in units of micro-Algos per byte.
-   * Fee may fall to zero but transactions must still have a fee of
-   * at least MinTxnFee for the current network protocol.
+   * Fee may fall to zero but a group of N atomic transactions must
+   * still have a fee of at least N*MinTxnFee for the current network protocol.
    */
   public fee: number | bigint;
 
@@ -1717,8 +1717,8 @@ export class TransactionParametersResponse extends BaseModel {
    * as of LastRound.
    * @param fee - Fee is the suggested transaction fee
    * Fee is in units of micro-Algos per byte.
-   * Fee may fall to zero but transactions must still have a fee of
-   * at least MinTxnFee for the current network protocol.
+   * Fee may fall to zero but a group of N atomic transactions must
+   * still have a fee of at least N*MinTxnFee for the current network protocol.
    * @param genesisHash - GenesisHash is the hash of the genesis block.
    * @param genesisId - GenesisID is an ID listed in the genesis block.
    * @param lastRound - LastRound indicates the last round seen
--- a/src/client/v2/algod/models/types.ts
+++ b/src/client/v2/algod/models/types.ts
@@ -1685,8 +1685,8 @@ export class TransactionParametersResponse extends BaseModel {
   /**
    * Fee is the suggested transaction fee
    * Fee is in units of micro-Algos per byte.
-   * Fee may fall to zero but a group of N atomic transactions must
-   * still have a fee of at least N*MinTxnFee for the current network protocol.
+   * Fee may fall to zero but transactions must still have a fee of
+   * at least MinTxnFee for the current network protocol.
    */
   public fee: number | bigint;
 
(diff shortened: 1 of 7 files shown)


src/makeTxn.ts
```diff
@@ -35,7 +35,7 @@ import { RenameProperties, RenameProperty } from './types/utils';
  * @param suggestedParams - a dict holding common-to-all-txns args:
  * fee - integer fee per byte, in microAlgos. for a flat fee, set flatFee to true
  * flatFee - bool optionally set this to true to specify fee as microalgos-per-txn
- *      If the final calculated fee is lower than the protocol minimum fee, the fee will be increased to match the minimum.
+ *       If true, txn fee may fall below the ALGORAND_MIN_TX_FEE for the current network protocol
```
jdtzmn, line 38: I'm not sure how I feel about the `current network protocol` language here. I could be wrong, but I think wording in the SDKs is "timeless" for the most part, not referencing a specific network protocol, with the one exception being TEAL versioning. I'm not positive about this though – I'd welcome other opinions on this.

tests/cucumber/steps/steps.js
```diff
@@ -4421,6 +4446,18 @@ module.exports = function getSteps(options) {
     }
   );
 
+  Then('fee field is in txn', async function () {
+    const s = algosdk.decodeObj(this.stx);
+    const { txn } = s;
+    assert.strictEqual('fee' in txn, true);
```
jdtzmn, line 4452: Not worth changing this over, but I thought I would mention that the `myObj.hasOwnProperty('key')` is preferred in JavaScript. The `in` operator will match keys across the object's entire prototype chain whereas `hasOwnProperty()` only checks the object's own keys. It's not a big deal at all in this case, but it can sometimes cause hard-to-catch bugs. JavaScript is weird 😄

jdtzmn (approved):
Looks good! I just had one question about doc wording that I'd love others to weigh in on.

src/client/v2/algod/models/types.ts
```diff
@@ -1685,8 +1685,8 @@ export class TransactionParametersResponse extends BaseModel {
   /**
```
jasonpaulos, line 1685: Could you remove the changes to this file since it's generated from the algod API spec? If these are included, I think we'll get an autogenerated PR to change this file back.

We should change the descriptions for fees in algod instead.

src/makeTxn.ts
jasonpaulos: In this case I agree we should remove the reference to the current protocol, saying "may fall below ALGORAND_MIN_TX_FEE" should be sufficient.

tests/5.Transaction.js
```diff
@@ -85,6 +85,7 @@ describe('Sign', () => {
       from: '7ZUECA7HFLZTXENRV24SHLU4AVPUTMTTDUFUBNBD64C73F3UHRTHAIOF6Q',
       to: '7ZUECA7HFLZTXENRV24SHLU4AVPUTMTTDUFUBNBD64C73F3UHRTHAIOF6Q',
       fee: 0,
+      flatFee: false,
```
jasonpaulos, line 88: I believe flatFee is false by default. Is it necessary to add it here?

tests/5.Transaction.js
```diff
@@ -97,6 +98,40 @@ describe('Sign', () => {
...
+      lastRound: 61,
+      genesisHash: 'JgsgCaCTqIaLeVhyL6XlRu3n7Rfk2FxMeK+wRSaQ7dI=',
+      note: new Uint8Array([123, 12, 200]),
+    };
+    const txn = new algosdk.Transaction(o);
+    assert.equal(txn.fee, 10);
+    const txnEnc = txn.get_obj_for_encoding();
+    assert.equal(txnEnc.fee, 10);
+  });
+
```
jasonpaulos, line 134: Could you add one more unit test where fee=0 and flatFee isn't mentioned to ensure that the txn will still have a fee of 1000?

jasonpaulos (changes requested):

tests/5.Transaction.js
shiqizng: line removed.

tests/5.Transaction.js
shiqizng: this case is covered after removing line 88.

jasonpaulos: Good point, thanks

jasonpaulos (approved):
Pending tests, looks good

Issue #371, opened 2021-06-09
Export rest API types


amityadav0:
we are building https://github.com/scale-it/algo-builder, previously we were using [types-algosdk](https://github.com/scale-it/algo-builder/blob/master/packages/types-algosdk/types/index.d.ts) which was written by us, now we are trying to migrate to js-algorand-sdk types.

therefore we require `AssetParams`, `ApplicationParams`, `MultisigMetadata`, `Multisig`, `LogicSig`, `Kmd`, `Wallets`, `Compile`, `pendingTransactionInfo`, `Schema's` etc all types
but currently they are not exported in `types/index.d.ts`

is it possible to export all types so that we can use them ? 

Thanks

PR #391, opened 2021-06-25
Export Rest API types

amityadav0:
#371


src/types/transactions/index.ts
```diff
@@ -23,7 +23,38 @@ export {
...
+  PendingTransactionResponse,
+  TransactionParametersResponse,
+  DryrunRequest,
+  DryrunSource,
+  CompileResponse,
+  NodeStatusResponse,
+  TealKeyValue,
+  TealValue,
+  Account as AccountState,
+} from '../../client/v2/algod/models/types';
```
jasonpaulos, line 57: These models are already exported under `algosdk.modelsv2`.

jasonpaulos (changes requested):
Thank you for making this PR. We would be happy to accept a change that exports the additional encoded types, but the types for models have already been exported under `algosdk.modelsv2`.

src/types/transactions/index.ts
amityadav0: ok, thanks, removing these from here.

amityadav0:
Thanks, removed `algosdk.modelsv2`

src/types/transactions/index.ts
```diff
@@ -16,14 +16,26 @@ import {
...
+  EncodedSignedTransaction,
+  EncodedAssetParams,
+  EncodedGlobalStateSchema,
+  EncodedLocalStateSchema,
+  EncodedLogicSig,
+  EncodedMultisig,
+  EncodedMultisigBlob,
+  EncodedSubsig,
+} from './encoded';
+export { MultisigMetadata } from '../multisig';
```
jasonpaulos, line 37: Instead of modifying this file, could you create a new file `src/types/index.ts` that exports the types from `src/types/multisig.ts`, `src/types/address.ts`, and `src/types/transactions/index.ts` (probably with `export * from './transactions'`)?

I'd like to avoid importing files from a parent directly for the sole purpose of exporting things from them.

src/types/transactions/index.ts
```diff
@@ -16,14 +16,26 @@ import {
   ApplicationClearStateTransaction as AppClearStateTxn,
   ApplicationNoOpTransaction as AppNoOpTxn,
 } from './application';
-
+import SuggestedParamsRequest from '../../client/v2/algod/suggestedParams';
```
jasonpaulos, line 19: Why is this export needed?

jasonpaulos (changes requested):

src/types/transactions/index.ts
amityadav0: yes

src/types/transactions/index.ts
amityadav0: testfg

this type is being used while testing, a stub is used to override `getTransactionParams` function

jasonpaulos: I would recommend using `ReturnType<algosdk.Algodv2['getTransactionParams']>` instead of exporting `SuggestedParamsRequest` if this is the only reason, since if we export `SuggestedParamsRequest`, we should probably export all of the internal request builder types, and that is a larger task that I'm not sure the library should do at this point.

jasonpaulos:
Thank you for the update! This is almost ready, the last issue is that I would rather not export `SuggestedParamsRequest` at this time. I left a comment explaining a workaround using `ReturnType<*>`.

src/types/transactions/index.ts
amityadav0: okay, thanks. this works

amityadav0: removed.

jasonpaulos (approved):
This looks good, thanks

src/types/transactions/index.ts
```diff
@@ -16,14 +16,13 @@ import {
   ApplicationClearStateTransaction as AppClearStateTxn,
   ApplicationNoOpTransaction as AppNoOpTxn,
 } from './application';
-
```
jasonpaulos, line 19: One minor thing: could you add back this removed line? I can merge once that's done.

amityadav0: done

PR #392, opened 2021-06-25
Fix: Cast hexadecimal literals to BigInt to be compatible with React Native applications on android

lassmann:
Hey guys! I'm part of Rand Labs team and we are creating an algo-wallet for React Native.
The sdk works well on iOS but on Android throws the next error:
image

Doing a debug, this is because of the way of comparing a big int directly with hexadecimal

I make a quick fix on this PR jus casting the hexadecimal to a BigInt


jasonpaulos (approved):
Thanks for creating this. I have ran into the issue before and I'm glad you took the time to fix it.

Provided the tests pass, this looks good to merge.

PR #393, opened 2021-06-26
replacing truthy check with undefined

barnjamin:
I'm not sure if this is something that needs to be changed here or in AlgoSigner to set the decimal to an optional value.  Setting it to 0 currently fails validation on AlgoSigner because it is deleted here.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -594,7 +594,7 @@ export class Transaction implements TransactionStorageStructure {
         delete txn.apar;
       } else {
         if (!txn.apar.t) delete txn.apar.t;
-        if (!txn.apar.dc) delete txn.apar.dc;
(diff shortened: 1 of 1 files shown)


jasonpaulos:
Due to the way asset parameters are encoded by algod (specifically, [this struct](https://github.com/algorand/go-algorand/blob/abc4058b52be80c99e1ddfe9ff6d7156fc50f3d1/data/basics/userBalance.go#L332)), if you want to use a value of 0 for the decimals field, that field should be omitted/undefined. This is part of a larger space saving technique where any zero-value ints, strings, arrays, etc. are omitted from the canonical encoding.

So the fix should probably be that AlgoSigner needs to interpret an undefined `txn.apar.dc` field as being 0.

PR #397, opened 2021-07-07
Support transactions with a first round of zero

jdtzmn:
This PR fixes a bug that incorrectly prohibited encoding a transaction with a first round of zero. When using the new devMode flag, rounds are incremented by transactions instead of the flow of time and so it is often the case that the last round (and therefore the first valid round) received from the transaction params endpoint is 0.

To support that use case, this PR adds the first valid round property to the list of transaction properties that are allowed to be zero.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,6 +28,10 @@ module.exports = {
       'always',
       { exceptAfterSingleLine: true },
     ],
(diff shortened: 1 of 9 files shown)


jasonpaulos (approved):
Pending tests, looks good

PR #399, opened 2021-07-08
Fix msgpack encoding of dryrun objects

### LINKED ISSUE #373: DryRunResponse for stateful NoOp call
I am trying to produce a `DryRunRequest` and get it's response. I followed the [tutorial](https://developer.algorand.org/docs/features/asc1/debugging/#using-dryrun-for-debugging-a-teal-program) for lsig, which works fine, but having an issue with the same for stateful call (`NoOp`) request.

### Steps
```js
const tx = algosdk.makeApplicationNoOpTxn(..);
const app = algodClient.getApplicationByID(app_id).do(); // deployed stateful application id
const dr = new algosdk.modelsv2.DryrunRequest({
  apps: [ app ]
  txns: [{ txn: tx }],
  sources: undefined
}):
const resp = await algodClient.dryrun(dr).do();
```
here is the response:
```
{
  "error": "",
  "protocol-version": "https://github.com/algorandfoundation/specs/tree/ac2255d586c4474d4ebcf3809acccb59b7ef34ff",
  "txns": [
    {
      "app-call-messages": [
        "ApprovalProgram",
        "REJECT",
        "logic eval error: program version 65 greater than max supported version 3"
      ],
      "app-call-trace": [
        {
          "error": "program version 65 greater than max supported version 3",
          "line": 0,
          "pc": 0,
          "stack": []
        }
      ],
      "disassembly": null
    }
  ]
}
```
Here are the smart contracts:
- *approval.teal*
  ```
  #pragma version 3
  
  int NoOp
  txn OnCompletion
  ==

  bnz application_call

  int 1
  return

  application_call:

  // read global state
  byte "counter"
  dup
  app_global_get

  // increment the value
  int 1
  +

  // store to scratch space
  dup
  store 0

  // update global state
  app_global_put

  // load return value as approval
  load 0
  return
  ```
  
- *clear.teal*
  ```
  #pragma version 3
  int 1
  return
  ```

jasonpaulos:
This fixes an issue where the `ApplicationParams` dryrun model object expects programs to be base64 strings, but these programs were not properly encoded to msgpack for use with algod's dryrun endpoint.

Now, programs can be passed as `Uint8Array`s or base64 strings. Internally any base64 strings will be decoded to `Uint8Array`s.

I've added an argument to `BaseModel.get_obj_for_encoding` called `binary`, which indicates whether raw binary objects (`Uint8Array`s) are acceptable or if they should be encoded as base64 strings.

Fixes algorand/go-algorand#2495, fixes #373.

### DIFF
--- a/src/client/v2/algod/models/base.ts
+++ b/src/client/v2/algod/models/base.ts
@@ -3,8 +3,8 @@
  */
 
 /* eslint-disable no-underscore-dangle,camelcase */
-
 function _is_primitive(val: any): val is string | boolean | number | bigint {
+  /* eslint-enable no-underscore-dangle,camelcase */
   return (
     val === undefined ||
     val == null ||
@@ -12,7 +12,7 @@ function _is_primitive(val: any): val is string | boolean | number | bigint {
   );
 }
 
-/* eslint-disable no-redeclare,no-unused-vars */
+/* eslint-disable no-underscore-dangle,camelcase,no-redeclare,no-unused-vars */
 function _get_obj_for_encoding(
   val: Function,
   binary: boolean
@@ -23,6 +23,7 @@ function _get_obj_for_encoding(
   binary: boolean
 ): Record<string, any>;
 function _get_obj_for_encoding(val: any, binary: boolean): any {
+  /* eslint-enable no-underscore-dangle,camelcase,no-redeclare,no-unused-vars */
   let targetPropValue: any;
 
   if (val instanceof Uint8Array) {
@@ -47,9 +48,9 @@ function _get_obj_for_encoding(val: any, binary: boolean): any {
   }
   return targetPropValue;
 }
-/* eslint-enable no-redeclare,no-unused-vars */
 
 export default class BaseModel {
+  /* eslint-disable no-underscore-dangle,camelcase */
   attribute_map: Record<string, string>;
 
   /**
@@ -60,6 +61,7 @@ export default class BaseModel {
    *   be used for objects that will be encoded with JSON.
    */
   get_obj_for_encoding(binary = false) {
+    /* eslint-enable no-underscore-dangle,camelcase */
     const obj: Record<string, any> = {};
 
     for (const prop of Object.keys(this.attribute_map)) {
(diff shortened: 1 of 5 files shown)


src/client/v2/algod/models/base.ts
```diff
@@ -1,63 +1,71 @@
-import { Address } from '../../../../types/address';
-
 /**
  * Base class for models
  */
 
-/* eslint-disable no-underscore-dangle,camelcase,class-methods-use-this */
-export default class BaseModel {
-  attribute_map: Record<string, string>;
+/* eslint-disable no-underscore-dangle,camelcase */
```
jdtzmn, line 5: ESLint disable markers should be re-enabled after the disabled code block. I think these dangling `eslint-disables` are actually my fault, but we might as well add them now. Additionally, I'm not sure that disabling `no-redeclare` and `no-unused-vars` is required now that we're using the [`@typescript-eslint/parser` versions of those rules](https://github.com/algorand/js-algorand-sdk/blob/develop/.eslintrc.js#L31-L34).

src/client/v2/algod/models/base.ts
```diff
@@ -1,63 +1,71 @@
...
-  /* eslint-disable no-dupe-class-members */
+  return targetPropValue;
+}
+/* eslint-disable no-redeclare */
+
+export default class BaseModel {
+  attribute_map: Record<string, string>;
 
-  get_obj_for_encoding() {
+  get_obj_for_encoding(binary: boolean = false) {
```
jdtzmn, line 63: optional change: boolean is implicit here due to the default

jdtzmn (changes requested):
Looks good, just a few comments. Have you tested this change by sending TEAL 4 programs to the dryrun endpoint?

src/client/v2/algod/models/base.ts
jasonpaulos: My latest commit re-enabled the rules. However, I found that my eslint still required them to be there.

jdtzmn (approved):
Looks good!

PR #400, opened 2021-07-08
Add `prepare` script to support npm installing with the git URL

jdtzmn:
This PR adds a `prepare` script to the `package.json` to support npm installing this repository with the git URL like so:

```sh
$ npm install --save https://github.com/algorand/js-algorand-sdk
```

The above command worked previously when the `dist` directory was stored in version control but, since it has been removed, the convention [starting in NPM v4](https://stackoverflow.com/questions/48287776/automatically-build-npm-module-on-install-from-github/57503862#57503862) is to add any necessary build scripts to the `prepare` script in `package.json`, so that npm knows how to build your cloned repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
   },
   "scripts": {
     "test": "node -r ts-node/register tests/mocha.js",
+    "prepare": "npm run build",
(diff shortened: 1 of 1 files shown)


jasonpaulos (approved):
Tested with `npm i --save git://github.com/algorand/js-algorand-sdk#jacob/prepare-script`, looks good to me 🚀

Issue #389, opened 2021-06-24
`EncodedSignedTransaction` is missing `msig` field


amityadav0:
`EncodedSignedTransaction` should have `msig`, because while decoding transaction using algosdk.decodeSignedTransaction(stxn); `msig is lost`

jasonpaulos:
This appears to be just a type issue, meaning if you call `algosdk.decodeSignedTransaction` with a multisig transaction, it will have an `msig` property but TypeScript doesn't think it does. @amityadav0 let me know if this is not the case.

Also, the `lsig` field is also missing.

amityadav0:
yes, this is the case. it's a type issue

jasonpaulos:
Closed by #401.

PR #401, opened 2021-07-12
add `msig` and `lsig` fields to `EncodedSignedTransaction` type

amityadav0:
#389

### DIFF
--- a/src/types/transactions/encoded.ts
+++ b/src/types/transactions/encoded.ts
@@ -293,26 +293,6 @@ export interface EncodedTransaction {
   apep?: number;
 }
 
-/**
- * A structure for an encoded signed transaction object
- */
-export interface EncodedSignedTransaction {
-  /**
-   * Transaction signature
-   */
-  sig: Buffer;
-
-  /**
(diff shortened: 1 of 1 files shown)


src/types/transactions/encoded.ts
```diff
@@ -360,3 +340,33 @@ export interface EncodedLogicSig {
...
 }
+
+/**
+ * A structure for an encoded signed transaction object
+ */
+export interface EncodedSignedTransaction {
+  /**
+   * Transaction signature
+   */
+  sig: Buffer;
```
jasonpaulos, line 351:
```suggestion
  sig?: Buffer;
```
This field should be optional too. Only one of `sig`, `lsig`, or `msig` can be defined at once.

jasonpaulos (changes requested):
Thanks for making this. Just one change needs to be made now.

src/types/transactions/encoded.ts
amityadav0: got it.

jasonpaulos (approved):

PR #402, opened 2021-07-13
Migrate multisig tests to TypeScript and fix various type issues

### LINKED ISSUE #378: Incorrect return type for `decodeSignedTransaction`
The `decodeSignedTransaction` method returns a type that is supposed to have an encoded transaction object, but it actually has an instance of the `Transaction` class, as seen in the attached implementation.

https://github.com/algorand/js-algorand-sdk/blob/fd187cedab70cb123333ba62ca3e6fe98c1e7d3d/src/transaction.ts#L1064-L1070

To solve this, a new type `SignedTransaction` should be created to differentiate between a signed transaction object with an encoded transaction and a signed transaction object with a transaction instance.

jasonpaulos:
This PR migrates the MultiSig unit test file to TypeScript and fixes some type issues that came up along the way. Specifically:
* `Transaction.from_obj_for_encoding` used to have a return type of `any`, not it's `Transaction`.
* Created a new type `SignedTransaction` for the return type of `decodeSignedTransaction`.
* Made the fields `fee`, `fv`, `note`, `lx` of `EncodedTransaction` optional, since they can be dropped if 0 or empty.

Closes #378


src/transaction.ts
```diff
@@ -1073,18 +1075,53 @@ export function decodeUnsignedTransaction(
   return Transaction.from_obj_for_encoding(partlyDecodedObject);
 }
 
+/**
+ * Object representing a transaction with a signature
+ */
+export interface SignedTransaction {
```
jdtzmn, line 1081: This should be moved to the `types` directory. Perhaps `src/types/transactions/base.ts` would make sense?

jdtzmn (changes requested):
Looks good, just had one comment.

src/transaction.ts
jasonpaulos: Unfortunately placing this type elsewhere would create a dependency cycle, since it needs to import `Transaction`. We could move `decodeSignedTransaction` to a different file to break this loop though, what do you think about that?

jdtzmn: That's a tough one. I'm hesitant to relocate only that function to a new file, but it feels bad to store a widely-used type somewhere other than the `types` directory. Maybe the solution is to leave it here for now and relocate it later once we have a "transaction auxiliary functions" file when a sufficient number of functions similar to `decodeSignedTransaction` have been added.

jdtzmn (approved):

mxmauro:
Hi @jasonpaulos , why the `export` of `createMultisigTransaction` was removed? It was useful to create templates for multisig TXs without the signature.

It is very useful to have those functions exposed in order to be able to create TXs and adding the hardware wallet signatures.

Regards,
Mauro.

jasonpaulos:
@mxmauro I stopped exporting `createMultisigTransaction` because it was meant to be an internal function used during a call to `signMultisigTransaction`. Also, I don't believe it was ever exported by the top-level `algosdk` package, so I did not expect this to affect any user code.

How are you currently importing this function, and can you use `signMultisigTransaction` as a replacement?

mxmauro:
Hi @jasonpaulos , don't worry, I know it isn't the best approach to use something not on the top for JS libraries.

We would like to have some functionallity like `createMultisigTransaction` so we have the whole multisig TX object and then we manually add the signature.

This is because you don't have access to the account's `sk` when you use hardware wallets.

jasonpaulos:
@mxmauro so the problem is that you have no way of creating the initial unsigned multisig transaction bytes if the SDK does not know the secret keys of any of the signers?

If that's the case, the SDK should definitely provide a solution to that.

Alex99y:
Yes, @jasonpaulos ! Right now, we are using `createMultisigTransaction` to create the EncodedMultisigTransaction using the signature provided by the hardware wallet when the user signs the transaction. Would be nice if AlgoSDK could export some functions to create MultisigTransaction using any of EncodedTransaction/Transaction and a Signature with the address associated with it.

jasonpaulos:
Thank you for the additional context. This seems like a shortcoming of the current SDK functions. I'll make an issue to better support this use case.

PR #408, opened 2021-07-21
Signing support for rekeying to LogicSig/MultiSig account

### LINKED ISSUE #365: Allow 'null' args for makeLogicSig
It seems makeLogicSig was changed to where it only checks undefined args, although it used to check for null args. For backwards compatibility purposes it should be checking both. 

For example this code fails from the makeLogicSig line returning null, due to the null args. The source code should still compile even if the args are null. The current workaround of course is to change it to undefined.

The fix is that makeLogicSig should accept both undefined and null args, and still compile the source code even with null/undefined args.
```

        let compilation = await compileProgram(algodClient, program);
        let uintAr = _base64ToArrayBuffer(compilation.result);
        let args = null;
        let lsig = algosdk.makeLogicSig(uintAr, args);
        console.log("lsig addr: " + lsig.address());
```
        
The issue happens here: https://github.com/algorand/js-algorand-sdk/blob/fd187cedab70cb123333ba62ca3e6fe98c1e7d3d/src/logicsig.ts#L25

```


let args: Uint8Array[] | undefined;    
if (typeof bufferOrUint8ArrArgs !== 'undefined')      
args = bufferOrUint8ArrArgs.map((arg) => new Uint8Array(arg));


```

jasonpaulos:
Implement the improvements from https://github.com/algorand/go-algorand-sdk/pull/230

This also closes #365

### DIFF
--- a/src/logicsig.ts
+++ b/src/logicsig.ts
@@ -55,6 +55,8 @@ export class LogicSig implements LogicSigStorageStructure {
 
     this.logic = program;
     this.args = args;
+    this.sig = undefined;
+    this.msig = undefined;
   }
 
   // eslint-disable-next-line camelcase
@@ -76,8 +78,8 @@ export class LogicSig implements LogicSigStorageStructure {
   // eslint-disable-next-line camelcase
   static from_obj_for_encoding(encoded: EncodedLogicSig) {
     const lsig = new LogicSig(encoded.l, encoded.arg);
-    if (encoded.sig) lsig.sig = encoded.sig;
-    if (encoded.msig) lsig.msig = encoded.msig;
+    lsig.sig = encoded.sig;
+    lsig.msig = encoded.msig;
     return lsig;
   }
 
@@ -210,6 +212,7 @@ export class LogicSigAccount {
    */
   constructor(program: Uint8Array, args?: Array<Uint8Array | Buffer> | null) {
     this.lsig = new LogicSig(program, args);
+    this.sigkey = undefined;
   }
 
   // eslint-disable-next-line camelcase
@@ -227,7 +230,7 @@ export class LogicSigAccount {
   static from_obj_for_encoding(encoded: EncodedLogicSigAccount) {
     const lsigAccount = new LogicSigAccount(encoded.lsig.l, encoded.lsig.arg);
     lsigAccount.lsig = LogicSig.from_obj_for_encoding(encoded.lsig);
-    if (encoded.sigkey) lsigAccount.sigkey = encoded.sigkey;
+    lsigAccount.sigkey = encoded.sigkey;
     return lsigAccount;
   }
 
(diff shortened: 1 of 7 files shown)


tests/8.LogicSig.ts
```diff
@@ -0,0 +1,1487 @@
...
+            genesisHash: 'JgsgCaCTqIaLeVhyL6XlRu3n7Rfk2FxMeK+wRSaQ7dI=',
+          },
+          note: new Uint8Array([180, 81, 121, 57, 252, 250, 210, 113]),
+        });
+
+        assert.throws(
+          () => algosdk.signLogicSigTransaction(txn, lsig),
+          (err) =>
+            err.message ===
+            'Logic signature verification failed. Ensure the program and signature are valid.'
```
jdtzmn, line 547: If only checking for equality, the following `assert.throws` overload is more concise:

```
assert.throws(() => algosdk.signLogicSigTransaction(txn, lsig), 'Logic signature verification failed. Ensure the program and signature are valid.')
```

jdtzmn (approved):
Just two small changes, but otherwise looks good! Feel free to merge once they're fixed.

src/logicsig.ts
```diff
@@ -193,9 +194,162 @@ export class LogicSig implements LogicSigStorageStructure {
   }
 }
 
+/**
+ * Represents an account that can sign with a LogicSig program.
+ */
+export class LogicSigAccount {
```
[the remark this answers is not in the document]
jasonpaulos: That's a good idea, maybe an `Encodable<T>` class, where `T` is the type of the encoded object.

src/main.ts
```diff
@@ -260,6 +253,7 @@ export {
 } from './convert';
 export { computeGroupID, assignGroupID } from './group';
 export {
+  LogicSigAccount,
```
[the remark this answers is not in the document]
jasonpaulos: It was actually never exported, only the `makeLogicSig` function was. Since it's depreciated I see no need to fix that now.

tests/8.LogicSig.ts
jasonpaulos: Good point, I originally thought that syntax wasn't valid but I think I had other mistakes that prevented it from working

src/main.ts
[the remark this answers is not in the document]
jdtzmn: Huh- I could've sworn I saw it exported. I'll resolve this.

tests/8.LogicSig.ts
jasonpaulos: Actually I still get this error when changing this to the above: `TypeError [ERR_AMBIGUOUS_ARGUMENT]: The "error/message" argument is ambiguous. The error message "Logic signature verification failed. Ensure the program and signature are valid." is identical to the message.`

jasonpaulos: According to https://nodejs.org/api/assert.html#assert_assert_throws_fn_error_message a string isn't allowed as the second argument. I could change it to a RegExp to get it to work, but it's not clear to me whether that would be better than the current code.

PR #409, opened 2021-07-26
export logicsig, update metadatahash type

amityadav0:
- Add string type in metadatahash, becuase in transaction types metadatahash can be string too.
- Export logicsig

- Update logicsig function
hk


src/logicsig.ts
```diff
@@ -128,7 +128,7 @@ export class LogicSig implements LogicSigStorageStructure {
    * @param msig - Multisig account as \{version, threshold, addrs\}
    */
   sign(secretKey: Uint8Array, msig?: MultisigMetadata) {
-    if (msig === undefined) {
```
Alex99y: It could be better: `if (!msig) { ... }`. `msig` will never be null, it can be `MultisigMetadata` or `undefined`.

robert-zaremba:
Can we have it merged? We will need it for the next release of `algo-builder`. Otherwise we will need to use our own fork. @jasonpaulos

src/client/v2/algod/models/types.ts
```diff
@@ -688,7 +688,7 @@ export class AssetParams extends BaseModel {
    * (am) A commitment to some unspecified asset metadata. The format of this
    * metadata is up to the application.
    */
-  public metadataHash?: Uint8Array;
+  public metadataHash?: Uint8Array | string;
```
jasonpaulos, line 691: What is the purpose of this change? Recently we changed the constructor of this type to allow `metadataHash` (and all other binary data fields) to accept a base64 string as well: https://github.com/algorand/js-algorand-sdk/blob/63c657018a112ad0f9a7e105e82b5157d96ed636/src/client/v2/algod/models/types.ts#L784

This file is actually generated from algod's REST spec, so it would be difficult to change the type of this specific field anyway.

src/main.ts
```diff
@@ -260,6 +260,7 @@ export {
   logicSigFromByte,
   tealSign,
   tealSignFromProgram,
+  LogicSig,
```
jasonpaulos, line 263: Also, what is the motivation behind exporting `LogicSig` now? It probably _should_ have been exported all along, but it's now deprecated in favor of `LogicSigAccount`, which is exported.

jasonpaulos (commented):
I'm uncertain these changes are needed, which I explain below. Please let me know if there are use cases I'm not aware of which require these changes.

src/main.ts
robert-zaremba: Ah, we missed that it's deprecated. We were using `LogicSig` type for a year and then when we migrated from our generated typescript typing to the new typescript SDK it was missing. We will try to migrate our codebase.

src/client/v2/algod/models/types.ts
amityadav0: we are using `AssetParams` type in runtime in which user can pass string type in metadatahash, it seems it will be difficult to change here, so we will add the conversion using recent changes`Recently we changed the constructor of this type to allow metadataHash (and all other binary data fields) to accept a base64 string`.

amityadav0:
Thanks for the reply, i think we don't need this PR, we will update our code according to recent changes made.

PR #423, opened 2021-08-20
Implemented WaitForConfirmation function

DamianB-BitFlipper:
When using the JS SDK, I oftentimes find myself copying the same code to WaitForConfirmation() of a pending transaction.

The WaitForConfirmation() waits for a pending transaction to be accepted and confirmed by the network. Until then, it blocks. I have implemented this function with a timeout feature as well that will error if the timeout is achieved.

I am unsure what to do about the CHANGELOG

This code was taken from the utils.js in the `examples` directory and moved into the actual v2 code of this SDK.

### DIFF
--- a/src/client/v2/algod/waitForConfirmation.ts
+++ b/src/client/v2/algod/waitForConfirmation.ts
@@ -6,14 +6,14 @@ import HTTPClient from '../../client';
  * confirmed by the network
  */
 export default class WaitForConfirmation extends JSONRequest {
-  constructor(c: HTTPClient, private txid: string, private timeout: number) {
+  constructor(c: HTTPClient, private txid: string, private waitRounds: number) {
     super(c);
     this.txid = txid;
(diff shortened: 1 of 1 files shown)


barnjamin:
I've got the same copy-pasta'd function in my project.  This would be a nice addition, alternatively a sendWait that blocks until the txn is confirmed

jasonpaulos (changes requested):
Similar to https://github.com/algorand/go-algorand-sdk/pull/232, can you implement this as a helper function, not as a REST endpoint on the algod client?

I would suggesting making a new file in the `src` folder just for this, modifying `main.ts` to export it, and using it in `tests/cucumber/steps/steps.js` to ensure it works.

jasonpaulos:
Thank you for working on this @DamianB-BitFlipper! I'm closing this because I've made #469 to continue development.

PR #428, opened 2021-08-31
remove contract template feature

brianolson:
part of https://github.com/algorand/go-algorand-internal/issues/1415


algobolson:
(this is passing except that one test is failing to download the latest Chrome from Google)

algobarb:
Just fyi, I just merged in a change to update chromedriver so it should pass if you pull in the latest change in package.json and package-lock.json.

michaeldiamant:
Closing - We intend to pick up work here in the near future, but we can do so with a fresh PR.  Thanks to @brianolson for providing a starting point.

PR #442, opened 2021-09-24
updating example client to v2

barnjamin:
Should this  be against develop branch?


jasonpaulos:
Yes, could you change the target branch to develop? We only push to master when we have a new release

PR #450, opened 2021-09-30
fix: add support for deserializing nonparticipating transaction

code-dev-ai:
`Transaction.from_obj_for_encoding` fails to process nonparticipating transactions without `voteKey`, `selectionKey` or `voteKeyDilution` fields which are not required for nonparticipating transactions.

This PR fixes the issue and voteKey`, `selectionKey` or `voteKeyDilution` are not required for nonparticipating transactions.

PR #451, opened 2021-09-30
Support deserializing nonpart and offline keyreg

fabrice102:
This PR extends
    https://github.com/algorand/js-algorand-sdk/pull/450
adding support for offline key registration transactions too.

Previously offline key registration would not be deserializable.


jasonpaulos (approved):
Thank you for this fix

PR #454, opened 2021-10-11
ABI Support for JS library

### LINKED ISSUE #425: Add ABI encoding support
## Summary
Add support for encoding and decoding ABI values, as described in https://github.com/algorandfoundation/ARCs/pull/7

## Scope
Add ability to:
* Parse and serialize ABI type strings, e.g. `(uint64, byte[], address)`
* Encode and decode ABI values, using type information

algochoi:
This PR adds ABI type encoding support for the JS SDK library.

Closes #425


src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+import algosdk from '../..';
+import { concatArrays } from '../utils/utils';
+
+export const MAX_LEN = 2 ** 16 - 1;
+export const ADDR_BYTE_SIZE = 32;
+export const SINGLE_BYTE_SIZE = 1;
+export const SINGLE_BOOL_SIZE = 1;
+export const LENGTH_ENCODE_BYTE_SIZE = 2;
+
+export interface Segment {
```
jasonpaulos, line 25: This probably should not be exported

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+    // | ufixed <N> x <M> (8 <= N <= 512, N % 8 = 0, and 0 < M <= 160)
+    // | bool
+    // | address (alias for byte[32])
+    // | <type> [<N>]
+    // | <type> []
+    // | string
+    // | (T1, ..., Tn)
+*/
+import algosdk from '../..';
+import { concatArrays } from '../utils/utils';
```
jasonpaulos, line 17: To avoid circular imports, you should do `import { encodeAddress, decodeAddress } from '../encoding/address'` instead

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+  // Checks if the Type object (or any of its child types) have dynamic length
+  abstract IsDynamic(): boolean;
+  // Returns the size of the Type object in bytes
+  abstract ByteLen(): number;
+  // Encodes a value for the Type object using the ABI specs
+  abstract Encode(value: ABIValue): Uint8Array;
+  // Decodes a value for the Type object using the ABI specs
+  abstract Decode(byteString: Uint8Array): ABIValue;
+  // De-serializes the ABI type from a string using the ABI specs
+  static Of(str: String): Type {
```
jasonpaulos, line 49: `string` instead of `String` is preferred for this type

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+    if (str.endsWith('[]')) {
+      const arrayArgType = Type.Of(str.slice(0, str.length - 2));
+      if (arrayArgType === null) {
```
jasonpaulos, line 52: I don't think the return type will ever be null, will it?

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+      return new AddressType();
+    }
+    if (str === 'string') {
+      return new StringType();
+    }
+    if (str.length >= 2 && str[0] === '(' && str[str.length - 1] === ')') {
+      const tupleContent = TupleType.parseTupleContent(
+        str.slice(1, str.length - 1)
+      );
+      if (tupleContent.length === 0) {
```
jasonpaulos, line 112: nit: this special case does not actually change anything

src/abi/abi_type.ts
```diff
@@ -0,0 +1,649 @@
...
+
+  ByteLen() {
+    return this.bitSize / 8;
+  }
+
+  Encode(value: BigInt) {
+    if (
+      typeof value !== 'bigint' ||
+      value >= BigInt(2 ** this.bitSize) ||
+      value < 0n
```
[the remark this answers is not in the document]
jasonpaulos: Yes it's the same, but `0n` is not supported everywhere yet.

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+    if (
+      (typeof value !== 'bigint' && typeof value !== 'number') ||
+      value >= BigInt(2 ** this.bitSize) ||
+      value < 0n
+    ) {
+      throw new Error(
+        `${value} is not a non-negative int or too big to fit in size uint${this.bitSize}`
+      );
+    }
+    if (typeof value === 'number' && value >= 2 ** 53) {
```
jasonpaulos, line 164: You should use `Number.isSafeInteger(value)` instead. As a bonus, this will also error on decimal numbers

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+    if (
+      (typeof value !== 'bigint' && typeof value !== 'number') ||
+      value >= BigInt(2 ** this.bitSize) ||
+      value < 0n
+    ) {
+      throw new Error(
+        `${value} is not a non-negative int or too big to fit in size ufixed${this.bitSize}`
+      );
+    }
+    if (typeof value === 'number' && value >= 2 ** 53) {
```
jasonpaulos, line 225: Same thing here

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+    }
+    return new Uint8Array([0]);
+  }
+
+  Decode(byteString: Uint8Array) {
+    if (byteString.length !== 1) {
+      throw new Error(`bool string must be 1 byte long`);
+    }
+    const buf = Buffer.from(byteString);
+    const value = buf.readUIntBE(0, byteString.length);
```
jasonpaulos, line 307: nit: this could be simplified to `const value = byteString[0];`

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+    }
+    return new Uint8Array([value]);
+  }
+
+  Decode(byteString: Uint8Array) {
+    if (byteString.length !== 1) {
+      throw new Error(`byte string must be 1 byte long`);
+    }
+    const buf = Buffer.from(byteString);
+    return buf.readUIntBE(0, SINGLE_BYTE_SIZE);
```
jasonpaulos, line 347: Same thing here

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+
+    // Concatenate into fixed Uint8Array
+    let resultBytes = new Uint8Array();
+    for (const h of heads) {
+      resultBytes = concatArrays(resultBytes, h);
+    }
+    for (const t of tails) {
+      resultBytes = concatArrays(resultBytes, t);
+    }
+    return resultBytes;
```
jasonpaulos, line 625: nit: `return concatArrays(...heads, ...tails)` would be simpler

src/abi/abi_type.ts
```diff
@@ -0,0 +1,649 @@
...
+// Convert a BigInt to a big-endian Uint8Array for encoding.
+function bigIntToBytes(bi: BigInt | number, size: number) {
+  let hex = bi.toString(16);
+  // Pad the hex with zeros so it matches the size in bytes
+  for (let i = hex.length; i < size * 2; i++) {
+    hex = `0${hex}`;
+  }
+  const byteArray = new Uint8Array(hex.length / 2);
+  for (let i = 0, j = 0; i < hex.length / 2; i++, j += 2) {
+    byteArray[i] = parseInt(hex.slice(j, j + 2), 16);
```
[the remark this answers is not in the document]
jasonpaulos: Weirdly this seems to be the best way, so I'll resolve this

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+        }
+      } else {
+        until -= 1;
+        break;
+      }
+    }
+    return until;
+  }
+
+  static parseTupleContent(str: String): String[] {
```
jasonpaulos, line 766: Should be `string` and `string[]`

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+        const ti = Type.Of(tupleContent[i]);
+        tupleTypes.push(ti);
+      }
+      return new TupleType(tupleTypes);
+    }
+    throw new Error(`cannot convert a string ${str} to an ABI type`);
+  }
+}
+
+export class UintType implements Type {
```
jasonpaulos, line 126: I think these subclasses should have `extend Type` instead of `implements Type`

package.json
```diff
@@ -64,6 +64,9 @@
     "webpack": "^5.38.1",
     "webpack-cli": "^4.7.2"
   },
+  "browser": {
+    "crypto": false
```
jasonpaulos, line 68: What is the purpose of this? Right now I think it may be causing some problems, since the key `browser` is also defined before this

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+      tupleStrings.push(word);
+    }
+    if (depth !== 0) {
+      throw new Error('tuple string has mismatched parentheses');
+    }
+    return tupleStrings;
+  }
+
+  // compressMultipleBool compresses consecutive bool values into a byte in ABI tuple / array value.
+  static compressMultipleBool(valueList: ABIValue[]): number {
```
jasonpaulos, line 805: nit: functions `compressMultipleBool` and `findBoolLR` aren't really related to `TupleType`, so I'd suggest making them their own standalone functions instead of static functions on that class.

src/abi/abi_type.ts
```diff
@@ -0,0 +1,846 @@
...
+
+type ABIValue = boolean | number | bigint | string | Uint8Array | ABIValue[];
+
+export abstract class Type {
+  // Converts a Type object to a string
+  abstract toString(): string;
+  // Checks if two Type objects are equal in value
+  abstract equal(other: Type): boolean;
+  // Checks if the Type object (or any of its child types) have dynamic length
+  abstract IsDynamic(): boolean;
```
jasonpaulos, line 41: JS function names aren't usually capitalized, so I'd suggest making them lowercase. E.g. `IsDynamic => isDynamic, ByteLen => byteLen, etc.`

jasonpaulos (changes requested):
I have the following file organization suggestions, what do you think?
* Move the functions  `bigIntToBytes` and `bytesToBigInt` to a new file `src/encoding/bigint.ts`, since they are useful in general
* Right now all ABI code is in a folder by itself. Since we can't split up the file easily without introducing circular imports, I think moving the file to `src/encoding/abi.ts` would make more sense

Also, some exporting suggestions:
* I just realized if we export the code as is, users would do `import { Type } from 'algosdk'`, which is not great because `Type` is pretty general. Probably renaming to `ABIType` makes more sense.
* `src/main.ts` needs to be updated to actually export the new class. If you adopt the above changes, `export { ABIType } from './encoding/abi';` should do it

(somehow half my review got submitted early, that was not intentional)

package.json
ahangsu: This is an ugly fix for the status quo, if we do not disable the `crypto` field, then the Travis test in browser, and local test in browser, will both fail.

We are not quite sure of a clear solution. I believe the current fix comes from: https://stackoverflow.com/questions/54162297/module-not-found-error-cant-resolve-crypto/54162447.

Personally both Andrew and I do not like duplicate the same field in json config, any thoughts on that?

src/abi/abi_type.ts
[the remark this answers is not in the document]
algochoi: Gotcha, thanks!

src/abi/abi_type.ts
algochoi: Right, good catch. It should throw an error if there is nothing before it anyway, i.e. `[]`

src/abi/abi_type.ts
algochoi: Neat!

package.json
jasonpaulos: This appears to be an issue with the test script, not with the build in general, since `npm run build` still works.

I think the actually issue is a circular dependency somehow. I made the change from https://github.com/algorand/js-algorand-sdk/pull/454#discussion_r742418470 and changed the import line in `10.ABI.ts` from `import algosdk from '..';` to `import algosdk from '../index';`, and now `npm test` works properly now without this `package.json` change. @algochoi let me know if this works for you as well.

algochoi: I think it works locally too -- I changed all the `import algosdk` dependencies and just pushed the changes 😃

jasonpaulos:
Weirdly it looks like the latest commit is failing on travis because of new cucumber tests from https://github.com/algorand/algorand-sdk-testing/pull/145, but that shouldn't be happening because they're hidden behind the `@unit.transactions.keyreg` tag...

src/encoding/bigint.ts
```diff
@@ -0,0 +1,33 @@
...
+ * bytesToBigInt produces a bigint from a binary representation.
+ *
+ * @param bytes - The Uint8Array to convert.
+ * @returns The bigint that was encoded in the input data.
+ */
+export function bytesToBigInt(bytes: Uint8Array) {
+  let res = 0n;
+  const buf = Buffer.from(bytes);
+  for (let i = 0; i < bytes.length; i++) {
+    res = BigInt(Number(buf.readUIntBE(i, 1))) + res * 256n;
```
jasonpaulos, line 30:
```suggestion
  let res = BigInt(0);
  const buf = Buffer.from(bytes);
  for (let i = 0; i < bytes.length; i++) {
    res = BigInt(Number(buf.readUIntBE(i, 1))) + res * BigInt(256);
```

jasonpaulos (approved):
Looks good, except there are still some bigint literals 😄 

After fixing, this looks good to merge to me!

PR #463, opened 2021-11-03
Add stateproof keyreg field

jasonpaulos:
Add stateproof key field to keyreg transaction, in line with the changes from https://github.com/algorand/go-algorand/pull/2990


src/makeTxn.ts
```diff
@@ -192,7 +195,8 @@ export function makeKeyRegistrationTxnWithSuggestedParams(
   voteKeyDilution: any,
   suggestedParams: any,
   rekeyTo?: any,
-  nonParticipation = false
+  nonParticipation = false,
```
shiqizng, line 198: nit: does this need to have a default value? it looks like that value is always set in the other constructors.

src/makeTxn.ts
```diff
@@ -206,6 +210,7 @@ export function makeKeyRegistrationTxnWithSuggestedParams(
     type: TransactionType.keyreg,
```
shiqizng, line 210: you didn't make these changes, but do you know why type and reKeyTo are are labeled but the others are not?

src/transaction.ts
```diff
@@ -448,6 +489,7 @@ export class Transaction implements TransactionStorageStructure {
       !txn.nonParticipation &&
       (txn.voteKey ||
```
shiqizng, line 490: why is this check needed here. It's checking one of the fields is set?

shiqizng (commented):
it looks good to me, just very minor comments.

src/transaction.ts
jasonpaulos: This is to make sure if any of the online keyreg fields are provided, then all must be provided

src/makeTxn.ts
jasonpaulos: `type` is specified because it should always be `keyreg`, and `reKeyTo` is specified because unfortunately it's defined as `reKeyTo` and `rekeyTo` (notice the difference in the k) in different places, so we need to convert it.

src/makeTxn.ts
jasonpaulos: This default value has runtime implications (i.e. JS will fill in `false` if no value is provided when the function is called) while the other definitions are purely for static type analysis

jasonpaulos (commented):
Looks good. Though once stateproofs are merged into master, I would like to do a quick review again

winder (approved):

PR #466, opened 2021-11-09
Add ABI interaction support

### LINKED ISSUE #457: ABI Interaction - off-chain programs invoking on-chain contracts
As described in [this design document](https://gist.github.com/jasonpaulos/a810abe7e86d43840d14445718565a9a), we build support in the SDK for off-chain programs invoking on-chain contracts.

jasonpaulos:
To do:
- [x] Add Method, Interface, and Contract objects
- [x] Add `AtomicTransactionComposer`
- [x] Implement JSON description cucumber tests
- [x] Implement `AtomicTransactionComposer` cucumber tests

Closes #457

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -186,6 +186,11 @@ export class AtomicTransactionComposer {
       );
     }
 
+    if (methodArgs == null) {
+      // eslint-disable-next-line no-param-reassign
+      methodArgs = [];
+    }
+
     if (methodArgs.length !== method.args.length) {
       throw new Error(
         `Incorrect number of method arguments. Expected ${method.args.length}, got ${methodArgs.length}`
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -139,7 +139,7 @@ export class AtomicTransactionComposer {
    * for the given method.
    */
   addMethodCall({
-    appID,
+    appId,
     method,
     methodArgs,
     sender,
@@ -151,7 +151,7 @@ export class AtomicTransactionComposer {
     signer,
   }: {
     /** The ID of the smart contract to call */
-    appID: number;
+    appId: number;
     /** The method to call on the smart contract */
     method: ABIMethod;
     /** The arguments to include in the method call. If omitted, no arguments will be passed to the method. */
@@ -245,14 +245,14 @@ export class AtomicTransactionComposer {
       appArgsEncoded.push(appArgTypes[i].encode(appArgValues[i]));
     }
 
-    if (appID === 0) {
+    if (appId === 0) {
       throw new Error('Application create call not supported');
     }
 
     const appCall = {
       txn: makeApplicationCallTxnFromObject({
         from: sender,
-        appIndex: appID,
+        appIndex: appId,
         appArgs: appArgsEncoded,
         onComplete:
           onComplete == null ? OnApplicationComplete.NoOpOC : onComplete,
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -1,0 +1,480 @@
+import {
+  ABIType,
+  ABITupleType,
+  ABIValue,
+  ABIMethod,
+  abiTypeIsTransaction,
+} from './abi';
+import { decodeSignedTransaction } from './transaction';
+import { makeApplicationCallTxnFromObject } from './makeTxn';
+import { assignGroupID } from './group';
+import { waitForConfirmation } from './wait';
+import Algodv2 from './client/v2/algod/algod';
+import {
+  TransactionSigner,
+  TransactionWithSigner,
+  isTransactionWithSigner,
+} from './signer';
+import {
+  OnApplicationComplete,
+  SuggestedParams,
+} from './types/transactions/base';
+
+export type ABIMethodArgument = ABIValue | TransactionWithSigner;
+
+/** Represents the output from a successful ABI method call. */
+export interface ABIResult {
+  /** The TxID of the transaction that invoked the ABI method call. */
+  txID: string;
+  /**
+   * The raw bytes of the return value from the ABI method call. This will be empty if the method
+   * does not return a value (return type "void").
+   */
+  rawReturnValue: Uint8Array;
+  /**
+   * The return value from the ABI method call. This will be undefined if the method does not return
+   * a value (return type "void"), or if the SDK was unable to decode the returned value.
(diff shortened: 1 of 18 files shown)


algochoi (commented):
Nice and clean implementation -- just minor comments

src/composer.ts
```diff
@@ -0,0 +1,491 @@
...
+      // not quite a deep copy, but good enough for our purposes (modifying txn.group in buildGroup)
+      txn: Transaction.from_obj_for_encoding({
+        ...txn.get_obj_for_encoding(),
+        // erase the group ID
+        grp: undefined,
+      }),
+      signer,
+    }));
+    theClone.methodCalls = new Map(this.methodCalls);
+    theClone.signedTxns = this.signedTxns.slice();
```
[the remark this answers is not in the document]
jasonpaulos: Good point, copying `signedTxns` and `txIDs` is pointless since the copy is reverting to the BUILDING stage

src/composer.ts
```diff
@@ -0,0 +1,491 @@
...
+      );
+    }
+
+    if (this.transactions.length === AtomicTransactionComposer.MAX_GROUP_SIZE) {
+      throw new Error(
+        `Adding an additional transaction exceeds the maximum atomic group size of ${AtomicTransactionComposer.MAX_GROUP_SIZE}`
+      );
+    }
+
+    if (txnAndSigner.txn.group && txnAndSigner.txn.group.some((v) => v !== 0)) {
```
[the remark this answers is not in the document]
jasonpaulos: Yes, since `groupID = hash(hash(txn1) + hash(txn2) + ...))`, if any of the transactions already has a group ID, its hash will be taken with that group ID. So after we replace that with the generated group ID, the whole group won't verify properly anymore.

tests/cucumber/steps/steps.js
```diff
@@ -4476,6 +4551,384 @@ module.exports = function getSteps(options) {
...
+      );
+      assert.ok(this.composerExecuteResponse.confirmedRound > 0);
+    }
+  );
+
+  Then(
+    'The app should have returned {string}.',
+    function (base64ExpectedReturnValue) {
+      const { methodResults } = this.composerExecuteResponse;
+      assert.strictEqual(methodResults.length, 1);
```
algochoi, line 4903: non issue, just fyi: I think this could be generalized to checking multiple method returns in the future.

tests/cucumber/steps/steps.js
```diff
@@ -4476,6 +4551,384 @@ module.exports = function getSteps(options) {
...
+    'I add the current transaction with signer to the composer.',
+    function () {
+      this.composer.addTransaction(this.transactionWithSigner);
+    }
+  );
+
+  When(
+    'I build the transaction group with the composer. If there is an error it is {string}.',
+    function (errorMsg) {
+      assert.strictEqual(errorMsg, ''); // error checking not yet implemented
```
algochoi, line 4832: Good forward thinking

src/composer.ts
```diff
@@ -0,0 +1,497 @@
...
+
+          for (let i = logs.length - 1; i >= 0; i--) {
+            const log = Buffer.from(logs[i], 'base64');
+            if (log.byteLength >= 4 && log.slice(0, 4).equals(returnPrefix)) {
+              returnValueEncoded = log.slice(4);
+              break;
+            }
+          }
+
+          if (returnValueEncoded == null) {
```
algochoi, line 475: Interesting, is it better to do a loose check here? I think it's because you also want to check for empty slices in addition to the `undefined`, but I'm thinking what would happen for an ABI method with a valid four-byte-only return (maybe an empty tuple?).

algochoi (commented):
Looks good! I think we can merge this after one more pair of eyes looks at it.

tests/cucumber/steps/steps.js
jasonpaulos: Yes I saw the Python implementation does that, which makes a lot of sense. In the interest of getting this merged soon I'll leave it alone for now, but it would be nice to have a general version going forward.

src/composer.ts
jasonpaulos: This if statement will only execute if nothing in the logs array starts with the return prefix. If a method does return an "empty" value by logging only the 4 byte return hash, then `returnValueEncoded` would be a 0 length `Buffer` here, which is not `== null`.

algochoi (approved):

src/makeTxn.ts
```diff
@@ -1714,3 +1714,77 @@ export function makeApplicationNoOpTxnFromObject(
...
+    appAccounts: options.accounts,
+    appForeignApps: options.foreignApps,
+    appForeignAssets: options.foreignAssets,
+    note: options.note,
+    lease: options.lease,
+    reKeyTo: options.rekeyTo,
+    extraPages: options.extraPages,
+  };
+  return new txnBuilder.Transaction(o);
+}
```
ahangsu, line 1790: I am dumb and I know nothing about type-script, but this bunch of code is trying to define a function to build a txn from all these options? (given all the function input parameters?)

jasonpaulos: Yes that's right. This is needed because we only have functions to make app call transactions for specific OnComplete values, e.g. `makeApplicationNoOpTxnFromObject`, `makeApplicationClearStateTxnFromObject`, etc.

src/composer.ts
```diff
@@ -0,0 +1,497 @@
...
+    theClone.transactions = this.transactions.map(({ txn, signer }) => ({
+      // not quite a deep copy, but good enough for our purposes (modifying txn.group in buildGroup)
+      txn: Transaction.from_obj_for_encoding({
+        ...txn.get_obj_for_encoding(),
+        // erase the group ID
+        grp: undefined,
+      }),
+      signer,
+    }));
+    theClone.methodCalls = new Map(this.methodCalls);
```
ahangsu, line 101: It seems to me that you deep-copied the transactions, erased group-id and signed transactions (if they exists), and deep-copied the methodCalls (I am not sure on that one)?

jasonpaulos: That's correct, the Map constructor performs a deep copy (well, deep enough for our purposes)

ahangsu (approved):
lgtm

PR #478, opened 2021-12-05
Create dryrun

barnjamin:
Adding method createDryrun to construct DryrunRequest from client and array of SignedTransaction objects

### DIFF
--- a/src/dryrun.ts
+++ b/src/dryrun.ts
@@ -39,14 +39,16 @@ export async function createDryrun({
   const appInfos = [];
   const acctInfos = [];
 
-  const apps = [];
-  const assets = [];
-  const accts = [];
+  const apps: number[] = [];
+  const assets: number[] = [];
+  const accts: string[] = [];
+
   for (const t of txns) {
     if (t.txn.type === TransactionType.appl) {
-      accts.push(t.txn.from);
+      accts.push(encodeAddress(t.txn.from.publicKey));
 
-      if (t.txn.appAccounts) accts.push(...t.txn.appAccounts);
+      if (t.txn.appAccounts)
+        accts.push(...t.txn.appAccounts.map((a) => encodeAddress(a.publicKey)));
 
       if (t.txn.appForeignApps) apps.push(...t.txn.appForeignApps);
 
--- a/src/dryrun.ts
+++ b/src/dryrun.ts
@@ -12,6 +12,31 @@ import { encodeAddress } from './encoding/address';
 
 const defaultAppId = 1380011588;
 
+// When writing the DryrunRequest object as msgpack the output needs to be the byte arrays not b64 string
+interface AppParamsWithPrograms {
+  ['approval-program']: string | Uint8Array;
+  ['clear-state-program']: string | Uint8Array;
+}
+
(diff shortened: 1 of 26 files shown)


src/dryrun.ts
```diff
@@ -0,0 +1,128 @@
...
+          'base64'
+        );
+        ai.params['clear-state-program'] = Buffer.from(
+          appInfo.params['clear-state-program'],
+          'base64'
+        );
+        appInfos.push(ai);
+      });
+  }
+  await Promise.all(appPromises);
```
jasonpaulos, line 106: `appPromises` never gets anything added to it. Same thing for `acctPromises` below.

Note since these 3 stages are all independent, it would be better to just `await Promise.all([... assetPromises, ...appPromises, ...acctPromises])` at the end, instead of an await after each of the 3.

src/dryrun.ts
```diff
@@ -0,0 +1,128 @@
...
+ * @returns the DryrunRequest object constructed from the SignedTransactions passed
+ */
+export async function createDryrun(
+  client: AlgodClient,
+  txns: SignedTransaction[],
+  protocolVersion?: string,
+  latestTimestamp?: number | bigint,
+  round?: number | bigint,
+  sources?: DryrunSource[]
+): Promise<DryrunRequest> {
```
jasonpaulos, line 29: nit: since there are many arguments, some of which are optional, I'd probably suggest using an argument object instead, something like:

```typescript
createDryrun({
  client,
  txns,
  protocolVersion,
  latestTimestamp,
  round,
  sources,
}: {
  client: AlgodClient;
  txns: SignedTransaction[];
  protocolVersion?: string;
  latestTimestamp?: number | bigint;
  round?: number | bigint;
  sources?: DryrunSource[];
}): Promise<DryrunRequest>
```

src/dryrun.ts
```diff
@@ -0,0 +1,139 @@
...
+
+      if (t.txn.appAccounts) accts.push(...t.txn.appAccounts);
+
+      if (t.txn.appForeignApps) apps.push(...t.txn.appForeignApps);
+
+      if (t.txn.appForeignAssets) assets.push(...t.txn.appForeignAssets);
+
+      // Create application,
+      if (t.txn.appIndex === 0) {
+        const appId = 1;
```
ahangsu, line 55: nit: https://github.com/algorand/go-algorand-sdk/pull/265/files#r763433869

barnjamin:
While testing the output as msgpack, tealdbg uses the applications in the `created-apps` of an account if its available instead of the app itself. I'm not sure that is correct to do but it exposed the fact that I was passing the b64 string instead of the bytes from the program.

The latest commit includes logic to decode the approval/clear-state programs after we get them back from the rest calls but before adding them to the info arrays.

I'm not happy with how its done here, please offer a better way if you've got one

jasonpaulos (approved):

src/dryrun.ts
fionnachan: a bit late to the party but I believe none of these can be optional because they are required params for the `DryrunRequest` constructor and they don't have a default value in `createDryrun`

PR #480, opened 2021-12-07
Fix wait for confirmation function

jasonpaulos:
Fixes the following issue with the wait for confirmation function:
1. If the call to `pendingTransactionInformation` throws an error, that error should be ignored and the function should continue retrying if `waitRounds` allows it.

### DIFF
--- a/src/wait.ts
+++ b/src/wait.ts
@@ -26,24 +26,27 @@ export async function waitForConfirmation(
 
   /* eslint-disable no-await-in-loop */
   while (currentRound < startRound + waitRounds) {
-    let pendingInfo = null;
+    let poolError = false;
     try {
-      pendingInfo = await client.pendingTransactionInformation(txid).do();
-    } catch (err) {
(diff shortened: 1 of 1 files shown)


src/wait.ts
```diff
@@ -26,16 +26,25 @@ export async function waitForConfirmation(
...
+    if (pendingInfo != null) {
+      if (pendingInfo['confirmed-round']) {
+        // Got the completed Transaction
+        return pendingInfo;
+      }
+
+      if (pendingInfo['pool-error']) {
+        // If there was a pool error, then the transaction has been rejected!
+        throw new Error(`Transaction Rejected: ${pendingInfo['pool-error']}`);
+      }
```
winder, line 47: Move into the try block.

You check `pendingInfo != null` but above we check `typeof status === 'undefined'`. Which of these is more correct? Maybe checking for null isn't needed inside the try block.

jasonpaulos: Good points, I'll modify this section

winder (approved):

PR #482, opened 2021-12-09
Support ABI reference types and other improvements

### LINKED ISSUE #468: Support foreign objects as ABI arguments
## Summary
This is a follow up to #466, which added support for invoking ABI methods.

In addition to standard ABI type arguments, the SDK should support the special foreign object argument types `account`, `asset`, and `application`, [as described in ARC-4](https://github.com/algorandfoundation/ARCs/blob/6a9f91aed8068bbace872483a64cbf22d0e1c975/ARCs/arc-0004.md#foreign-apps-asas-and-accounts).

## Scope
Extend the `AtomicTransactionComposer.addMethod` function to support these new method argument types. At a high level it should:

1. Recognize when an argument is a special type or a standard ABI type. The special argument types are only valid as top-level arguments, so direct string comparison should be sufficient.
2. If a special type is present, insert the argument value into the appropriate foreign array, and use the array index of the value as the app call argument, treated as an ABI `uint8`.

### LINKED ISSUE #483: Implement ABI changes
Implement the changes described in https://github.com/algorandfoundation/ARCs/pull/53

Additionally, the following implementation details should change:
* The atomic transaction composer should not add a group ID if there's only 1 transaction in the group, since it's unnecessary
* The atomic transaction composer should check that the type of a transaction argument (e.g. `pay`, `axfer`, etc.) matches the ABI transaction argument type.

jasonpaulos:
- [x] Add reference type support
- [x] Throw error if transaction type argument does not match transaction argument type
- [x] Add more type checking for implementations of `ABIType.encode` and `.equal`
- [x] AtomicTransactionComposer does not assign group ID if there's only 1 transaction in the group
- [x] Changes from https://github.com/algorandfoundation/ARCs/pull/53
  - [x] Don't tuple last argument when length is 15
  - [x] Modify `ABIContract` class
- [x] Add Contract and Interface descriptions, from  https://github.com/algorandfoundation/ARCs/pull/55
- [x] Changes from https://github.com/algorandfoundation/ARCs/pull/57
  - [x] AtomicTransactionComposer only checks the last log for a return value
  - [x] App creation and update transactions are supported in ABI method calls
- [x] Change `algorand-sdk-testing` branch back to `master` after https://github.com/algorand/algorand-sdk-testing/pull/153 is merged

Closes #468, closes #483

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -62,6 +62,19 @@ export enum AtomicTransactionComposerStatus {
   COMMITTED,
 }
 
+/**
+ * Add a value to an application call's foreign array. The addition will be as compact as possible,
+ * and this function will return an index that can be used to reference `valueToAdd` in `array`.
+ *
+ * @param valueToAdd - The value to add to the array. If this value is already present in the array,
+ *   it will not be added again. Instead, the existing index will be returned.
+ * @param array - The existing foreign array. This input may be modified to append `valueToAdd`.
+ * @param zeroValue - If provided, this value indicated two things: the 0 value is special for this
+ *   array, so all indexes into `array` must start at 1; additionally, if `valueToAdd` equals
+ *   `zeroValue`, then `valueToAdd` will not be added to the array, and instead the 0 indexes will
+ *   be returned.
+ * @returns An index that can be used to reference `valueToAdd` in `array`.
+ */
 function populateForeignArray<Type>(
   valueToAdd: Type,
   array: Type[],
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -25,6 +25,9 @@ import {
   SuggestedParams,
 } from './types/transactions/base';
 
+// first 4 bytes of SHA-512/256 hash of "return"
+const RETURN_PREFIX = Buffer.from([21, 31, 124, 117]);
+
 export type ABIArgument = ABIValue | TransactionWithSigner;
 
 /** Represents the output from a successful ABI method call. */
@@ -635,25 +638,19 @@ export class AtomicTransactionComposer {
                 await client.pendingTransactionInformation(txID).do();
 
           const logs: string[] = pendingInfo.logs || [];
-
-          // first 4 bytes of SHA-512/256 hash of "return"
-          const returnPrefix = Buffer.from([21, 31, 124, 117]);
-
-          let returnValueEncoded: Buffer | undefined;
-
-          for (let i = logs.length - 1; i >= 0; i--) {
-            const log = Buffer.from(logs[i], 'base64');
-            if (log.byteLength >= 4 && log.slice(0, 4).equals(returnPrefix)) {
-              returnValueEncoded = log.slice(4);
-              break;
-            }
-          }
-
-          if (returnValueEncoded == null) {
+          if (logs.length === 0) {
             throw new Error('App call transaction did not log a return value');
           }
 
-          methodResult.rawReturnValue = new Uint8Array(returnValueEncoded);
+          const lastLog = Buffer.from(logs[logs.length - 1], 'base64');
+          if (
+            lastLog.byteLength < 4 ||
+            !lastLog.slice(0, 4).equals(RETURN_PREFIX)
+          ) {
+            throw new Error('App call transaction did not log a return value');
+          }
+
+          methodResult.rawReturnValue = new Uint8Array(lastLog.slice(4));
           methodResult.returnValue = method.returns.type.decode(
             methodResult.rawReturnValue
           );
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -1,9 +1,14 @@
 import {
   ABIType,
   ABITupleType,
+  ABIUintType,
+  ABIAddressType,
   ABIValue,
   ABIMethod,
+  ABIReferenceType,
   abiTypeIsTransaction,
+  abiCheckTransactionType,
+  abiTypeIsReference,
 } from './abi';
 import { Transaction, decodeSignedTransaction } from './transaction';
 import { makeApplicationCallTxnFromObject } from './makeTxn';
@@ -55,6 +60,27 @@ export enum AtomicTransactionComposerStatus {
 
   /** The atomic group has been finalized, signed, submitted, and successfully committed to a block. */
   COMMITTED,
+}
+
+function populateForeignArray<Type>(
+  valueToAdd: Type,
+  array: Type[],
+  zeroValue?: Type
+): number {
+  if (zeroValue != null && valueToAdd === zeroValue) {
+    return 0;
+  }
+
+  const offset = zeroValue == null ? 0 : 1;
+
+  for (let i = 0; i < array.length; i++) {
+    if (valueToAdd === array[i]) {
+      return i + offset;
+    }
+  }
+
+  array.push(valueToAdd);
+  return array.length - 1 + offset;
 }
 
 /** A class used to construct and execute atomic transaction groups */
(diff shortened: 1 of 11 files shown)


src/composer.ts
```diff
@@ -57,6 +62,27 @@ export enum AtomicTransactionComposerStatus {
   COMMITTED,
 }
 
+function populateForeignArray<Type>(
```
ahangsu, line 65: nit: would be great to add comment to explain what this function is doing, i.e., calculate the object index for foreign object array...
btw, it seems the name is somewhat misleading tho?

jasonpaulos: It's based on the name for this function from goal: https://github.com/algorand/go-algorand/blob/29a10dad66d0ed5b793a18967e8ee234788c834f/cmd/goal/application.go#L1091

src/composer.ts
```diff
@@ -232,33 +335,108 @@ export class AtomicTransactionComposer {
...
+        case ABIReferenceType.account: {
+          const addressType = new ABIAddressType();
+          const address = addressType.decode(addressType.encode(refValue));
+          resolved = populateForeignArray(address, foreignAccounts, sender);
+          break;
+        }
+        case ABIReferenceType.application: {
+          const uint64Type = new ABIUintType(64);
+          const refAppID = uint64Type.decode(uint64Type.encode(refValue));
+          if (refAppID > Number.MAX_SAFE_INTEGER) {
```
algochoi, line 376: Could this check be moved into `populateForeignArray` and have that method take `ABIType` as a parameter as well?

algochoi (commented):
I like the new type checks! Overall looks good -- just had a few questions about resolving foreign objects in the composer.

src/composer.ts
```diff
@@ -137,19 +179,26 @@ export class AtomicTransactionComposer {
    * for the given method.
    */
   addMethodCall({
```
algochoi, line 181: Also, do we want to let users pass in their own foreign object arrays to the method call?

src/composer.ts
```diff
@@ -232,33 +335,108 @@ export class AtomicTransactionComposer {
...
+      }
+
+      basicArgTypes.push(argType);
+      basicArgValues.push(argValue);
+    }
+
+    const resolvedRefIndexes: number[] = [];
+    const foreignAccounts: string[] = [];
+    const foreignApps: number[] = [];
+    const foreignAssets: number[] = [];
```
ahangsu, line 360: One question, do we want to constrain the number of foreign asset/account/app number?
I remembered in @algochoi py-sdk that we defined some number of constraints.
```python
    # The maximum number of foreign objects that can be associated per app
    FOREIGN_ACCOUNT_LIMIT = 4
    FOREIGN_ARRAY_LIMIT = 8
```
Is it needed here in a similar sense?

src/composer.ts
```diff
@@ -232,33 +335,108 @@ export class AtomicTransactionComposer {
...
+              `Expected safe integer for application value, got ${refAppID}`
+            );
+          }
+          resolved = populateForeignArray(Number(refAppID), foreignApps, appID);
+          break;
+        }
+        case ABIReferenceType.asset: {
+          const uint64Type = new ABIUintType(64);
+          const refAssetID = uint64Type.decode(uint64Type.encode(refValue));
+          if (refAssetID > Number.MAX_SAFE_INTEGER) {
```
ahangsu, line 387: separate minor question: what is the expected range of foreign asset/app id, do we have a expected upper/lower bound for that?

src/composer.ts
```diff
@@ -232,33 +335,108 @@ export class AtomicTransactionComposer {
...
-      const lastArgTupleTypes = appArgTypes.slice(14);
-      const lastArgTupleValues = appArgValues.slice(14);
+    if (basicArgTypes.length > 15) {
+      const lastArgTupleTypes = basicArgTypes.slice(14);
+      const lastArgTupleValues = basicArgValues.slice(14);
 
-      appArgTypes = appArgTypes.slice(0, 14);
-      appArgValues = appArgValues.slice(0, 14);
+      basicArgTypes = basicArgTypes.slice(0, 14);
+      basicArgValues = basicArgValues.slice(0, 14);
```
ahangsu, line 412: nit: consider using some constant `MAX_ABI_ARG_NUM = 15` than numeric literals

ahangsu (commented):
Looks largely good to me, just some minor polishing here and there

src/composer.ts
```diff
@@ -232,33 +335,108 @@ export class AtomicTransactionComposer {
...
+    const foreignAssets: number[] = [];
+    for (let i = 0; i < refArgTypes.length; i++) {
+      const refType = refArgTypes[i];
+      const refValue = refArgValues[i];
+      let resolved = 0;
+
+      switch (refType) {
+        case ABIReferenceType.account: {
+          const addressType = new ABIAddressType();
+          const address = addressType.decode(addressType.encode(refValue));
```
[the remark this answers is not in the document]
jasonpaulos: We need to do this because there are multiple ways to represent an ABI value in JS. E.g. an `address` could be the base32 string, or it could be a Uint8Array with 32 bytes. And a `uint64` could be a JS `number` or `bigint`. And since the output of the `decode` function is consistent, it's sufficient to perform a round trip encode + decode in order to "normalize" the value.

src/composer.ts
jasonpaulos: This is a good suggestion, I'll attempt this change now

src/composer.ts
jasonpaulos: We probably should, but I chose to ignore this in the cucumber tests and JS and Go SDK implementations in order to get to a release sooner.

I can make follow up issues for it, but since this is a less necessary feature, I don't think it should block a release.

Though if the Python or Java SDK have already implemented it, that's perfectly fine.

src/composer.ts
jasonpaulos: Technically the range is 0 to the max uint64 value, but in practice actual asset & app IDs will be relatively "low" since they are always equal to the transaction sequence number.

And unfortunately the JS SDK uses a `number`, not a `bigint`, to represent IDs everywhere, so it has an implicit maximum value of [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER). In v2 of the SDK we should change this to a `bigint` to be safe & more consistent with our other values.

src/composer.ts
jasonpaulos: My point of view is that these should not be checked in the SDKs because:
* These are consensus params that are liable to change in the future, and if they do we'll need to quickly update the SDKs with the new values
* If too many values are included for one of these arrays, algod _should_ provide a sufficiently clear error message when you submit the transaction, so not much is gained by raising the error before submission
* The existing application call transaction types in each SDK do _not_ currently check these limits, so I don't think we should do checking at the AtomicTransactionComposer level either

ahangsu: Got it. If my understanding is correct, we might leave this to be handled on the algod side (which is 1 or somewhat more levels lower than AtomicTransactionComposer).

src/composer.ts
jasonpaulos: I tried implementing this, but the change would need to have much more complicated type annotations, since:
* TypeScript knows that specific types like `ABIUintType` and `ABIAddressType` return `bigint` and `string` respectively, which makes type checking easier. But if we pass a generic `ABIType` instance to the function, then we lose information about the return type and can only assume it returns `ABIValue`.
* `ABIUintType` actually returns a `bigint`, but we invoke `populateForeignArray` with its generic `Type` set to `number` since the foreign ASA and app arrays need to be `number[]`. This causes a problem because we need to add extra code to convert the `bigint` to a `number` that should only run in this case.

I don't think the extra complexity is worth it in this case.

src/composer.ts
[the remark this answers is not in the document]
jasonpaulos: I am considering adding a `normalize` function to `ABIType` for this specific purpose

algochoi (approved):

ahangsu (approved):
LGTM

PR #485, opened 2021-12-13
Improve HTTP error messages

### LINKED ISSUE #462: Improve error message for network requests
## Summary
Currently network errors have a generic message that says "Bad Request". We should improve this to have more useful information.

As a workaround, the following code can be used to extract more information about the error:

```javascript
  try {
    // algod request
  } catch (err) {
    // If network request, display verbose error
    if (err.response) {
      throw new Error(
        `Network request unsuccessful. Received status ${err.response.status}: ${err.response.text}`
      );
    }

    // Otherwise throw the error normally
    throw err;
  }
```

## Scope
I believe this is a problem with superagent, our HTTP request library. More info, plus some workarounds we may want to adopt: https://github.com/visionmedia/superagent/issues/1074

jasonpaulos:
Improve the HTTP error messages that superagent returns.

Before this PR, an error would look like:

```
Error: Bad Request
```

After this PR, it contains the error message returned from the server, like this:

```
Error: Network request unsuccessful. Received status 400: TransactionPool.Remember: transaction QJALLV5B3QDLCOOMOBGBEORNUGBRNNG6DVQ2DN42XATI4P2WFSAA: logic eval error: + overflowed. Details: pc=316, opcodes=load 4
       btoi
       +
```

Closes #462

### DIFF
--- a/src/client/urlTokenBaseHTTPClient.ts
+++ b/src/client/urlTokenBaseHTTPClient.ts
@@ -85,6 +85,22 @@ export class URLTokenBaseHTTPClient implements BaseHTTPClient {
     return res;
   }
 
+  /**
+   * Make a superagent error more readable. For more info, see https://github.com/visionmedia/superagent/issues/1074
+   */
+  private static formatSuperagentError(err: any): Error {
+    if (err.response) {
+      try {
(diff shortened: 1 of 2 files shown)


algochoi (approved):
LGTM :shipit:

PR #486, opened 2021-12-15
Document Algodv2 methods

fionnachan:
smile with tear
I was looking at this repo's source code and cross-checking with the REST APIs doc and got super annoyed....

**Live Deployment**
https://js-algosdk.netlify.app/classes/algodv2

### DIFF
--- a/src/client/v2/algod/algod.ts
+++ b/src/client/v2/algod/algod.ts
@@ -100,10 +100,14 @@ export default class AlgodClient extends ServiceClient {
    *
    * #### Example
    * ```typescript
-   * const { txId } = await algodClient.sendRawTransaction(stxns).do();
+   * const { txId } = await algodClient.sendRawTransaction(signedTxns).do();
+   * const result = await waitForConfirmation(algodClient, txid, 3);
    * ```
    *
    * [Response data schema details](https://developer.algorand.org/docs/rest-apis/algod/v2/#post-v2transactions)
+   *
+   * @remarks
+   * Often used with {@linkcode waitForConfirmation}
    * @param stxOrStxs - Signed transactions
    * @category POST
    */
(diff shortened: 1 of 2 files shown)


src/client/v2/algod/algod.ts
```diff
@@ -43,123 +65,343 @@ export default class AlgodClient extends ServiceClient {
...
+   * ```typescript
+   * const pendingTxns = await algodClient.pendingTransactionsInformation().do();
+   * ```
+   *
+   * #### Example 2
+   * ```typescript
+   * const maxTxns = 5;
+   * const getPendingTxns = algodClient.pendingTransactionsInformation();
+   * getPendingTxns.max(maxTxns);
+   * const pendingTxns = await getPendingTxns.do();
```
jasonpaulos, line 193: nit: a simpler way to do this would be:
```typescript
const pendingTxns = await algodClient
    .pendingTransactionsInformation()
    .max(maxTxns)
    .do();
```

Also applies to `pendingTransactionByAddress`

src/client/v2/serviceClient.ts
```diff
@@ -66,6 +68,7 @@ export default abstract class ServiceClient {
   }
 
   /**
+   * @ignore
```
jasonpaulos, line 71: Why ignore this? It's a useful method

jasonpaulos (commented):
Thank you for the contribution, this is a great improvement! I just have a few comments

src/client/v2/serviceClient.ts
fionnachan: I thought developers using the sdk probably don't need to know about it

src/client/v2/algod/algod.ts
fionnachan: Oh right, done! btw `pendingTransactionByAddress` is returning `PendingTransaction`**s**`ByAddress` 🤔

src/client/v2/algod/algod.ts
```diff
@@ -24,11 +24,37 @@ import {
...
+   * #### Example
+   * ```typescript
+   * const token  = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+   * const server = "http://localhost";
+   * const port   = 4001;
+   * const algodClient = new algosdk.Algodv2(token, server, port);
+   * ```
+   * @remarks
+   * The above configuration is for a sandbox private network.
+   * For applications on production, you are encouraged to run your own node, or use an Algorand REST API provider with a dedicated API key.
```
fionnachan, line 52: Given the recent outage, I've changed the base server example from algoexplorer to sandbox and added the remarks. Can you see if this is okay? @jasonpaulos

jasonpaulos (approved):
Sorry for the delay, this looks good!

Issue #494, opened 2022-01-12
Remove Template Feature Related Cucumber Tests


tzaffi:
## Problem

The integration test `template.feature` has been removed, but various steps that only belong to this test are still present.

## Solution

Remove the code associated with `template.feature` and no other running tests. In particular, here is a list of steps that can be removed:

* **Given** _a split contract with ratio <ratn> to <ratd> and minimum payment <min_pay>_
* **When** _I fund the contract account_
* **When** _I send the split transactions_
* **Given** _an HTLC contract with hash preimage "<preimage>"_
* **When** _I fund the contract account_
* **When** _I claim the algos_
* **Given** _a periodic payment contract with withdrawing window <wd_window> and period <period>_
* **When** _I claim the periodic payment_
* **Given** _a limit order contract with parameters <ratn> <ratd> <min_trade>_
* **When** _I swap assets for algos_
* **_Given_** a dynamic fee contract with amount <amt>
* **_Given_** I send the dynamic fee transactions

## Dependencies

Testing PR https://github.com/algorand/algorand-sdk-testing/pull/156 should be merged before merging the SDK PR.

## Urgency

Low

PR #498, opened 2022-01-15
Implement C2C tests

### LINKED ISSUE #487: C2C feature testing and implementation
## Problem

We want to test C2C inner transaction call described in [c2c-feature-branch](https://github.com/algorand/go-algorand/tree/feature/contract-to-contract/data/transactions/logic#inner-transactions), and the discussion/design-doc is [here](https://docs.google.com/document/d/1uGM62VsN1F4aSTNg4t2q21JmuwS-OO-7bFY0VcaZcro/edit#).

We might want to have some deep-calls (a chain of inner-trans call) as testing method, which should be designed in SDK testing. The ideal result is that we can walk through inner transactions cascaded in tree-like structure.

algochoi:
- [x]  Add pendingTransactionInformation field to ABI methodResult
- [x]  Implement all c2c.feature tests
- [x]  Implement new test in applications.verified
- [x]  Implement new test in compile.feature
- [x]  Deprecate template tests - related PR [here](https://github.com/algorand/js-algorand-sdk/pull/428/files) and related issue: #494 
- [x]  Revert branch change in `run_docker.sh`

Closes #487 #494


tests/cucumber/browser/test.js
```diff
@@ -1,9 +1,8 @@
 const assert = require('assert');
-const sha256 = require('js-sha256');
```
tzaffi, line 2: was this a bug in the test implementation that you are fixing?

tzaffi (approved):
Left a non-blocking question. 

LGTM

tests/cucumber/browser/test.js
algochoi: It was an import used by the template tests so I just removed it in this PR

src/composer.ts
```diff
@@ -658,6 +659,7 @@ export class AtomicTransactionComposer {
             methodResult.rawReturnValue
           );
         }
+        methodResult.txInfo = pendingInfo;
```
jasonpaulos, line 662: Can you put this line immediately after `pendingInfo` is created? The reason being an error might be thrown between its current placement and the assignment of `pendingInfo`, meaning the caller would not receive `txInfo` when they should.

src/main.ts
```diff
@@ -2,7 +2,6 @@ import * as nacl from './nacl/naclWrappers';
 import * as address from './encoding/address';
 import * as encoding from './encoding/encoding';
 import * as txnBuilder from './transaction';
-import * as LogicTemplatesCommonJSExport from './logicTemplates';
```
jasonpaulos, line 5: Unfortunately we can't remove the template exports without it being a breaking change -- this is akin to deleting the templates.

jasonpaulos (changes requested):
Only a few small but important comments

jasonpaulos (approved):

PR #499, opened 2022-01-17
Add app creator to dryrun request

barnjamin:
When doing an appGlobalGetEx the values are extracted from the balance record of the app creator.  If the app creator is not in the balance records passed in dryrun request the eval will report  the application is not found


src/dryrun.ts
```diff
@@ -131,9 +132,12 @@ export async function createDryrun({
         .then((appInfo) => {
           const ai = decodePrograms(appInfo as AppWithAppParams);
           appInfos.push(ai);
+          accts.push(ai.params.creator);
+          accts.push(getApplicationAddress(appId));
```
jasonpaulos, line 136: I think this will already be in accounts because of line 105

jasonpaulos (approved):

Issue #500, opened 2022-01-18
Asset "total issuance" error when creating from SDK, not present when using goal


daoarchitect:
### Asset "total issuance" error when creating from SDK, not present when using goal

When creating an asset using `makeAssetCreateTxnWithSuggestedParamsFromObject` with a `total` value of `18446744073709551615` and a decimal value of `6` I receive a `Total asset issuance must be a positive number and smaller than 2^64-1` error. 

However, when creating an asset using `goal asset create --creator <creator-address> --decimals 6 --name "Test" --unitname "TST" --total 18446744073709551615 --asseturl "https://test.com"` I can successfully create the asset.

### Your environment

* Software version:
```
12884901889
3.0.1.stable [rel/stable] (commit #b619b940)
go-algorand is licensed with AGPLv3.0
source code available at https://github.com/algorand/go-algorand
```

* Node status if applicable: 
```
Last committed block: 305686
Time since last block: 1.4s
Sync Time: 0.0s
Last consensus protocol: https://github.com/algorandfoundation/specs/tree/bc36005dbd776e6d1eaf0c560619bb183215645c
Next consensus protocol: https://github.com/algorandfoundation/specs/tree/bc36005dbd776e6d1eaf0c560619bb183215645c
Round for next consensus protocol: 305687
Next consensus protocol supported: true
Last Catchpoint:
Genesis ID: sandnet-v1
Genesis hash: +NlhvhK2YSpLLfOcZdEquVVx9PBtHRY/WxGL6bvoRVw=
```
* Operating System details. 
MacOS Montery 12.0.1

### Steps to reproduce

1. Attempt to create an asset with the above parameters using the SDK.
2. Attempt to create an asset using `goal` with the same parameters

### Expected behaviour

That my experience is consistent across both the SDK and goal and that I can successfully create an asset with a total of this size.

### Actual behaviour

I receive the error `Total asset issuance must be a positive number and smaller than 2^64-1` when attempting to create the asset using the `makeAssetCreateTxnWithSuggestedParamsFromObject` function in the JS SDK. I can successfully create the asset with the same parameters using goal.

barnjamin:
@daoarchitect I think you can pass a bigint into the asset total, https://github.com/algorand/js-algorand-sdk/blob/fd187cedab70cb123333ba62ca3e6fe98c1e7d3d/src/transaction.ts#L131

js is limited to max_safe_int but you can wrap the number to make js not complain

[...]

PR #501, opened 2022-01-18
adding note to use bigint

barnjamin:
Adding message to help debug when amount/assetAmount is large as in #500


jasonpaulos (approved):
Looks good

PR #502, opened 2022-01-20
Fix JSON decoding

aldur:
Hey team,

I think I found a bug in how we decode json responses; essentially, we always pick the wrong parser.

Look at the following code snippet:

```typescript
  const assetDetails = await client
    .getAssetByID(ASAIdx)
    .setIntDecoding(algosdk.IntDecoding.BIGINT)
    .do();
  console.log(typeof assetDetails.params.total);
```

Before: `number`
Now: `bigint`

I also think that we should add a unit test for this. My JS skills are somehow limited, but I'd be happy to give it a shot if you point me in the right direction :)

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -140,9 +140,8 @@ export default class HTTPClient {
    * @param status Status of the response (used in case parseJSON fails)
    * @param jsonOptions Options object to use to decode JSON responses. See
    *   utils.parseJSON for the options available.
-   * @private
    */
-  private static parseJSON(
+  public static parseJSON(
     text: string,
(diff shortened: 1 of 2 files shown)


jasonpaulos:
@aldur thank you for catching this! I agree we should also have a unit test for this before merging.

Perhaps you could make the `HTTPClient.parseJSON` function public (in order to test it) and add a unit test in `tests/9.Client.ts`?

aldur:
Added unit tests :)

jasonpaulos (approved):
Looks good, thanks!

PR #507, opened 2022-01-27
Moving from travis to circleci

algojack:
* running prettier through npx now (local install instead of global)
* upgraded to ubuntu focal from bionic (since circleci doesn't have bionic machines)
* parallelized tests according to their type and by browser
* transferred and configured gh-pages deploy
* removed travis config

Tested in circleci, and gh-pages tested on another branch: https://github.com/algorand/js-algorand-sdk/commit/c738ef15517882db99105c5b759335d8f493b585


.circleci/config.yml
```diff
@@ -0,0 +1,124 @@
...
+      - run: npm run lint
+
+  prettier:
+    docker:
+      - image: << pipeline.parameters.ubuntu_docker >>
+    steps:
+      - checkout
+      - install_dependencies
+      - run: |
+          npm install prettier
```
jasonpaulos, line 53: This should not be needed since prettier should already be installed during `install_dependencies` -- in fact this may install a different version of the package

.circleci/config.yml
```diff
@@ -0,0 +1,124 @@
...
+          command: |
+            << parameters.sudo >> apt -y update && << parameters.sudo >> apt -y upgrade
+            << parameters.sudo >> apt -y install curl make git build-essential jq unzip
+      - node/install:
+          node-version: '12'
+      - run:
+          name: npm install
+          command: |
+            set -e
+            npm install
```
jasonpaulos, line 124: [`npm ci`](https://docs.npmjs.com/cli/v8/commands/npm-ci) should be used instead. Otherwise a different version of our dependencies might get installed

jasonpaulos (commented):
A few comments

.circleci/config.yml
algojack: updated ty

.circleci/config.yml
algojack: updated ty

jasonpaulos (approved):
Looks great!

onetechnical (approved):
This looks fine to me in my quick grok.

egieseke (approved):
Looks good.

PR #512, opened 2022-02-03
Add installation instructions for vite users

fionnachan:
Developers on webpack 5 or vite often face errors when they first installed the sdk.

I think adding some instructions on the README would be helpful.


jasonpaulos (commented):
This is useful information, but in an effort to keep the README minimal, we've created an `FAQ.md`, and I think that would be a better place for this. Though I think the README should still have a sentence that links to the appropriate parts of the FAQ.

PR #514, opened 2022-02-09
Dryrun stack printer

barnjamin:
No idea if this is horrible.

Instead of using the built in v2 types that are generated, defined a new class that is essentially a port of the Python one.

Still needs the logic for passing 0 to make stack padding match longest line but it seems to work


src/dryrun.ts
```diff
@@ -167,3 +167,159 @@ export async function createDryrun({
...
+
+class DryrunTrace {
+  trace: DryrunTraceLine[] = [];
+
+  constructor(t: Record<string, any>[]) {
+    if (t === undefined) return;
+    this.trace = t.map((line) => new DryrunTraceLine(line));
+  }
+
+  getTrace(): any[] {
```
jasonpaulos, line 220: This returns `Array<[number, number, string]>`?

Btw I would suggest making an object interface for that tuple instead of using `[number, number, string]`. Something like this would be much clearer:

```typescript
interface Trace { // or whatever name makes sense
    line: number,
    pc: number,
    stack: string,
}
```

src/dryrun.ts
```diff
@@ -167,3 +167,159 @@ export async function createDryrun({
...
+      return '';
+    return this.trace(this.logicSigTrace, this.logicSigDisassemly);
+  }
+}
+
+export class DryrunResult {
+  error: string = '';
+  protocolVersion: string = '';
+  txns: DryrunTransactionResult[] = [];
+  constructor(drrResp: Record<string, any>) {
```
jasonpaulos, line 318: All the constructors take `Record<string, any>`, but they could be more specific. It seems like every constructor accesses a deterministic set of keys, so you could type these as object interfaces instead.

`Record<string, any>` is a step up from `any`, but not a big step 😅

jasonpaulos (commented):
Generally looks good, but I left some comments on ways to improve the type definitions.

I assume this SDK will want to implement the cucumber test you're working on as well?

src/dryrun.ts
```diff
@@ -167,3 +171,293 @@ export async function createDryrun({
...
+    this.globalDelta = dtr['global-delta'];
+    this.cost = dtr.cost;
+    this.logicSigMessages = dtr['logic-sig-messages'];
+    this.logicSigDisassemly = dtr['logic-sig-disassembly'];
+    this.logs = dtr.logs;
+    this.appCallTrace = new DryrunTrace(dtr['app-call-trace']);
+    this.logicSigTrace = new DryrunTrace(dtr['logic-sig-trace']);
+  }
+
+  appCallRejected(): boolean {
```
algochoi, line 343: Are these functions used somewhere?

algochoi (commented):
Looks pretty good to me - just some nits and I'd probably run the formatter again + merge the develop branch to this.

src/dryrun.ts
barnjamin: Only convenience methods for the user to do something like iterate over txns and print trace for only those that failed

src/dryrun.ts
```diff
@@ -167,3 +171,293 @@ export async function createDryrun({
...
+}
+
+class DryrunTransactionResult {
+  disassembly: string[] = [];
+  appCallMessages: string[] | undefined = [];
+  localDeltas: AccountStateDelta[] | undefined = [];
+  globalDelta: EvalDeltaKeyValue[] | undefined = [];
+  cost: number | undefined = 0;
+  logicSigMessages: string[] | undefined = [];
+  logicSigDisassemly: string[] | undefined = [];
```
algochoi, line 311: noob question: is there a reason we distinguish the `logicSigDisassembly` (also check spelling) with the other `disassembly`? Or is it just because dryrun does this and we mirror it?

src/dryrun.ts
```diff
@@ -167,3 +171,293 @@ export async function createDryrun({
...
+}
+
+interface DryrunTransactionResultResponse {
+  disassembly: string[];
+  appCallMessages: string[] | undefined;
+  localDeltas: AccountStateDelta[] | undefined;
+  globalDelta: EvalDeltaKeyValue[] | undefined;
+  cost: number | undefined;
+  logicSigMessages: string[] | undefined;
+  logicSigDisassemly: string[] | undefined;
```
algochoi, line 240: nit: spelling
```suggestion
  logicSigDisassembly: string[] | undefined;
```

src/dryrun.ts
```diff
@@ -167,3 +171,293 @@ export async function createDryrun({
     sources,
   });
 }
+
+interface StackValueResponse {
+  type: number;
+  bytes: string;
+  uint: number;
+}
+class DryrunStackValue {
```
algochoi, line 180: nit: add extra new line here?

algochoi (commented):
Not an expert at dryrun - but looking at the cucumber tests and seeing that others are okay with this change, I don't have any concerns. Just minor additional nits, but I can give a soft approval after the tests pass

src/dryrun.ts
barnjamin: Initially the dryrun endpoint only returned the last disassembly it saw, so in the case of an app call signed with logic you'd only get back the apps disassembly. In order to keep backwards compat I added lsig disassembly to the endpoint and dupe it in disassembly on algod. So they may be the same or different but more correct than before :)

algochoi: Got it thanks! LGTM - just need to change the remaining `logicSigDisassemly` to `logicSigDisassembly` I think and I'll approve

algochoi (approved):

PR #524, opened 2022-02-21
Document all Indexer methods

fionnachan:
#### Added docs for
- [x] lookupAssetTransactions
- [x] searchForAssets
- [x] lookupAccountByID
- [x] lookupApplications
- [x] searchForApplications
- [x] lookupTransactionsByID
- [x] lookupBlock
- [x] searchForTransactions
- [x] searchAccounts
- [x] lookupApplicationLogs

https://indexer-method-docs--js-algosdk.netlify.app/classes/indexer


src/client/v2/indexer/lookupAssetTransactions.ts
```diff
@@ -4,121 +4,388 @@ import IntDecoding from '../../../types/intDecoding';
...
+   * @param round
+   * @category query
+   */
   maxRound(round: number) {
     this.query['max-round'] = round;
     return this;
   }
 
-  // asset ID to filter with, as int
-  assetID(id: number) {
```
fionnachan, line 62: I removed this one because the `assetID` is already required in the constructor
let me know if you want to add it back

src/client/v2/indexer/lookupAccountTransactions.ts
```diff
@@ -119,7 +119,7 @@ export default class LookupAccountTransactions extends JSONRequest {
    *        .txid(txId)
    *        .do();
    * ```
-   *
+   * @remarks Alternatively, use `indexerClient.lookupTransactionByID(txnId).do()`
```
fionnachan, line 122: A few queries seemed to perform the same request as another more straightforward method so I added remarks to them

fionnachan:
added docs for `exclude` and `creator` at https://github.com/algorand/js-algorand-sdk/pull/524/commits/182e10cc4ed8c79aebf450c6a7aef2cc779775a7
this PR should be ready now
let me know if you'd like any changes @jasonpaulos

src/client/v2/indexer/lookupAssetTransactions.ts
jasonpaulos: Nice find, I was not aware of this. Though to maintain backwards compatibility, can you keep this, but mark it as `@deprecated` and don't document it?

src/client/v2/indexer/searchAccounts.ts
```diff
@@ -1,66 +1,261 @@
...
     return this;
   }
 
-  // include accounts that use this spending key
+  /**
+   * Include accounts that use this spending key.
+   *
+   * #### Example
+   * ```typescript
+   * const authAddr = "spending key";
```
jasonpaulos, line 174: nit: could you make this an address for clarity?

src/client/v2/indexer/searchForApplications.ts
```diff
@@ -1,36 +1,123 @@
...
   }
 
-  // token for pagination
+  /**
+   * Specify the next page of results.
+   *
+   * #### Example
+   * ```typescript
+   * const maxResults = 20;
+   * const nextToken = "APA6C7C3NCANRPIBUWQOF7WSKLJMK6RPQUVFLLDV4U5WCQE4DEF26D4E3E";
```
jasonpaulos, line 66: This is not at all a requirement, but my suggestion for `nextToken` docs would be to show two requests, something like this, so it's more clear where the value for `nextToken` should come from:

```javascript
const maxResults = 20;
const appsPage1 = await indexerClient
        .searchForApplications()
        .limit(maxResults)
        .do();

const appsPage2 = await indexerClient
        .searchForApplications()
        .limit(maxResults)
        .nextToken(appsPage1["next-token"])
        .do();
```

jasonpaulos (commented):
Thanks for working on this! I only have a few comments

fionnachan:
@jasonpaulos Updated

jasonpaulos (approved):
Sorry for the delay, this looks good!

PR #535, opened 2022-03-04
Update chromedriver

jasonpaulos:
Update chromedriver dependency to fix Chrome tests.

Since we have to update this pretty frequently to match the latest Chrome release, I've also added some code to automatically install the latest chromedriver in our CI process.


algochoi (approved):
LGTM, minor comment below

1. I checked the [chromedriver](https://www.npmjs.com/package/chromedriver) package and there might be a new `99.0.0` version as of yesterday. Not sure if it makes a difference.

jasonpaulos:
Thanks, I'll update to that version

PR #549, opened 2022-03-23
Allow Uint8Arrays public keys for kmd signing

vividn:
The user gets the public key from `algosdk.decodeAddress(authAddress)` in the Uint8Array format, and the kmd client can already accept this how it is implemented. So it makes sense to allow this in the types for `signTransactionWithSpecificPublicKey` and `signMultisigTransaction`


jasonpaulos:
Sorry for the late reply. This looks like a good change, so if you can sign the CLA and pass our CI tests, I'll approve it.

vividn:
Thanks for the response @jasonpaulos. It looks like the CI is failing because of a hash mismatch when downloading a docker image :thinking:. is there a way to retrigger the CI? Maybe it will work if it is tried again

jasonpaulos:
I've rerun the failed tests, so hopefully it will pass now

jasonpaulos:
@vividn actually the chrome test is still failing because this branch uses an older version of the chromedriver package. Could you merge the develop into here? That should fix it

vividn:
My branch seems up to date with algorand:develop
image

jasonpaulos:
Interesting, you're right. I'll try to investigate further

jasonpaulos:
There was a problem in our develop branch which caused the test failure. Once https://github.com/algorand/js-algorand-sdk/pull/564 is merged (and we merge it into develop as well), could you again merge develop into this branch? Tests should pass once that happens

vividn:
Sure I can do that. Thanks @jasonpaulos 
Alternatively, I could change this PR base branch to be release/v1.16.0 and then get it merged in there. I notice there are already some corrections to types in client/kmd.ts, so it would fit well. What do you think?

jasonpaulos:
@vividn while that would work, we're already far along in the release and I don't want to delay it any longer. This will go out in the next release we make.

vividn:
Success :smiley:

jasonpaulos (approved):
Great, thanks for your patience! 😅

Issue #525, opened 2022-02-22
makeAssetTransferTxnWithSuggestedParams should allow bigints


jeapostrophe:
Specifically the `assetIndex` could be a bigint

michaeldiamant:
Summarizing outcomes from a group brainstorm:
* Philosophically, it feels most correct to change fields from `number` to `bigint` everywhere that go-algorand defines a `uint64` data type.  In practice, it causes a backwards incompatibility and adds _some_ friction to developer ergonomics (can't perform arithmetic with `number` _and_ `bigint`).
* Given that changing field types feels out of reach, we agreed it'd be reasonable to expand function types to accept `number | bigint` + perform _safe_ coercion to `number`.
* We confirmed via visual inspection that Transaction construction defends against unsafe conversions to `number` using  `Number.isSafeInteger`.  Here's an example check:  https://github.com/algorand/js-algorand-sdk/blob/894e2871c2c41500716074052e3ffdc3f409e099/src/transaction.ts#L402-L411

PR #552, opened 2022-04-02
Feat/allow bigint in src/client

vuvoth:
## Changes
- Add Numeric type = bigint | number
- Use Numeric in client/indexer for query data in src/client


vuvth:
This PR is solve one part of issue #525. I have the problem when try to allow bigint in args(decode, transaction type validate). 
I will try to resolve it in next PR.

PR #553, opened 2022-04-02
Return min-fee from suggested params

AustP:
This would be useful


magmel48:
is there any update?

Eric-Warehime:
Sorry this took a bit to get to. This has been added in #760

PR #557, opened 2022-04-25
pass a Uint8Array instead of a string for importing a key

barnjamin:
Believe this is just an oversight


jasonpaulos (approved):
Thanks for this

PR #561, opened 2022-04-26
adding foreign app addr to dryrun

barnjamin:
We should be including the app address for foreign apps


michaeldiamant (approved):
@barnjamin Based on my (limited) understanding, looks right.  I visually confirmed accounts are de-duplicated before usage.

PR #566, opened 2022-05-11
Generate updated client API code

algoidurovic:
Depends on: https://github.com/algorand/go-algorand/pull/3957


jasonpaulos (commented):
Looks good except `test-harness` should not be checked in.

(I care less about the TEAL disassembly in this PR because the model is not used. There are additional non-generated changes that need to be made to add that endpoint, but that's of course out of scope of the PR)

michaeldiamant (approved):

jasonpaulos (approved):

PR #567, opened 2022-05-13
Refresh generated sources

michaeldiamant:
Regenerates from go-algorand `master` to limit diffs introduced in https://github.com/algorand/js-algorand-sdk/pull/566.

PR #569, opened 2022-05-20
Update generated files

Eric-Warehime:
Updated generated files--using the version of generator in review in https://github.com/algorand/generator/pull/31

The diff represents changes to the alogd spec--because it seems like we don't use the indexer spec to generate any JS files.


michaeldiamant (approved):
@Eric-Warehime Looks good - Matches the previously generated diff (https://github.com/algorand/js-algorand-sdk/pull/567), which suggests https://github.com/algorand/generator/pull/31 did _not_ introduce regressions.

PR #570, opened 2022-05-25
Build: Add SDK code generation workflow

Eric-Warehime:
## Summary

Adds a github workflow which will call our code gen action.  
Depends on action being defined in  our generator... See https://github.com/algorand/generator/pull/32 for more details and testing done.

Runs on a cron job daily at 23:20. The Github docs say to try to avoid common times because there will be natural spikes of event processing happening then, so I've defined all of the jobs at an arbitrary time during the night.

It will run the code generator, commit to a branch if necessary, and open a PR for us to review. It will no-op if the branch w/ identical changes already exist.

### DIFF
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -1,0 +1,19 @@
+name: "SDK Code Generation"
+on:
+  schedule:
+    - cron: '20 23 * * *'
+permissions:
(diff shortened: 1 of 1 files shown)


algojack (approved):

winder (approved):
LGTM

PR #583, opened 2022-06-07
adding getMethodByName function to Contract and Interface

barnjamin:


src/abi/method.ts
```diff
@@ -160,3 +160,20 @@ export class ABIMethod {
     });
   }
 }
+
+export function getMethodByName(methods: ABIMethod[], name: string): ABIMethod {
```
algochoi, line 164: I think it would be nice if there was a guard statement that checks `methods` is not `null` or is `instance of ABIMethod[]` here

algochoi (commented):
Minor comment while we wait for firefox tests

src/abi/method.ts
barnjamin: added this

algochoi: nice, thanks! Looks good to me

Makefile
```diff
@@ -1,8 +1,8 @@
...
 	
 integration:
-	node_modules/.bin/cucumber-js --tags "@algod or @assets or @auction or @kmd or @send or @indexer or @rekey or @send.keyregtxn or @dryrun or @compile or @applications or @indexer.applications or @applications.verified or @indexer.231 or @abi or @c2c" tests/cucumber/features --require-module ts-node/register --require tests/cucumber/steps/index.js
+	node_modules/.bin/cucumber-js -b --fail-fast --tags "@algod or @assets or @auction or @kmd or @send or @indexer or @rekey or @send.keyregtxn or @dryrun or @compile or @applications or @indexer.applications or @applications.verified or @indexer.231 or @abi or @c2c" tests/cucumber/features --require-module ts-node/register --require tests/cucumber/steps/index.js
```
jasonpaulos, line 5: Did you mean to leave these flags here? If so, why not also apply to unit tests? Full backtrace (`-b`) makes sense, but I'm not sure if it would be a good thing to have CI fail fast or not -- in the case of multiple errors, it's nice if CI can tell you all of them, not just the first.

tests/cucumber/steps/index.js
```diff
@@ -26,12 +26,19 @@ if (browser) {
...
-    chromeOptions.addArguments('--no-sandbox', '--headless', '--disable-gpu');
-    firefoxOptions.addArguments('-headless');
+    chromeOptions = chromeOptions.addArguments(
+      'no-sandbox',
+      'disable-gpu',
+      'headless'
+    );
+    firefoxOptions = firefoxOptions
+      .setPreference('remote.active-protocols', 1) // Sets protocol to only WebDriver BiDi
+      .setPreference('network.http.network-changed.timeout', 30) // In Docker tests this apparently matters
```
jasonpaulos, line 40: Interesting, I'd expect setting `remote.active-protocols` to 2 (CDP) to be the solution here, since in Firefox 101 they changed the default from 2 to 3 (1|2).

Also, for `network.http.network-changed.timeout`, have you tried more values, or just 30?

barnjamin: to be fair 1 was the first one I tried, something about the combination is tripping us.

I did not try more settings for timeout, 30s seemed sufficient, the default is 5

Makefile
barnjamin: just removing these for now, but fail fast is a good one to use in the case of docker-test since otherwise it seems to hang the terminal if you try to quit after a failed test

jasonpaulos (approved):
Looks good! Thanks again for all the debugging work related to Firefox.

Once @algochoi agrees, this looks good to merge, and I hope this will be the last of the flaky Firefox tests 🙏

algochoi (commented):
Code changes LGTM

PR #590, opened 2022-06-21
Dev Tools: Source map decoder

barnjamin:
Implements a class to accept and decode a source map.

Also adds the option to request source map from the client from the `/v2/teal/compile` endpoint


src/logic/sourcemap.ts
```diff
@@ -0,0 +1,133 @@
+/* eslint-disable no-bitwise */
+
+// Adapted from https://github.com/Rich-Harris/vlq, thanks!
```
jasonpaulos, line 3: Why not use https://www.npmjs.com/package/vlq or https://www.npmjs.com/package/sourcemap-codec (I've not looked into their differences) as a dependency?

barnjamin: I didn't want to introduce another dependency but I will just use that if its ok?

src/logic/sourcemap.ts
```diff
@@ -0,0 +1,67 @@
...
+      }
+
+      if (!(lastLine in this.lineToPc)) this.lineToPc[lastLine] = [];
+
+      this.lineToPc[lastLine].push(pc);
+      this.pcToLine[pc] = lastLine;
+    }
+  }
+
+  getLineForPc(pc: number): number {
```
michaeldiamant, line 60: @barnjamin Friendly visual reminder to implement an analog of https://github.com/algorand/py-algorand-sdk/pull/353#discussion_r910177370 here.  

From my _very_ limited Typescript knowledge, it seems the return type can be `string | undefined` to encode optionality.

michaeldiamant: Sorry - I meant `number | undefined` in my prior note.

Makefile
```diff
@@ -1,5 +1,5 @@
...
+	node_modules/.bin/cucumber-js --tags "@unit.offline or @unit.algod or @unit.indexer or @unit.rekey or @unit.tealsign or @unit.dryrun or @unit.applications or @unit.responses or @unit.transactions or @unit.transactions.keyreg or @unit.transactions.payment or @unit.responses.231 or @unit.feetest or @unit.indexer.logs or @unit.abijson or @unit.abijson.byname or @unit.atomic_transaction_composer or @unit.responses.unlimited_assets or @unit.indexer.ledger_refactoring or @unit.algod.ledger_refactoring or @unit.dryrun.trace.application or @unit.sourcemap.pctoline" tests/cucumber/features --require-module ts-node/register --require tests/cucumber/steps/index.js
 	
 integration:
```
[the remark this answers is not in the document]
michaeldiamant: @barnjamin Apologies if I'm mistakenly unresolving the conversation.  Should I expect to see `@compile.sourcemap` added to `integration` target?

src/logic/sourcemap.ts
```diff
@@ -0,0 +1,67 @@
...
+      throw new Error(`Only version 3 is supported, got ${this.version}`);
+
+    if (this.mappings === undefined)
+      throw new Error(
+        'mapping undefined, cannot build source map without `mapping`'
+      );
+
+    const pcList = this.mappings.split(';').map((m) => {
+      const decoded = vlq.decode(m);
+      if (decoded.length > 1) return decoded[2];
```
michaeldiamant, line 38: @barnjamin I'm probably missing some intuition, can you double check me?

The highlighted line retrieves the element at index 2.  So, I'd expect the length check to be > 2, but it reads > 1.

barnjamin: you're right indeed. 

I think in this case, it happens to be true that if the length > 1 it'll have the 2th element populated but that makes no sense to check. Updated with better check.

tests/cucumber/steps/steps.js
```diff
@@ -5341,6 +5341,58 @@ module.exports = function getSteps(options) {
...
+    this.sourcemap = new algosdk.SourceMap(js);
+  });
+
+  Then(
+    'the string composed of pc:line number equals {string}',
+    function (mapping) {
+      const buff = [];
+      for (const [pc, line] of Object.entries(this.sourcemap.pcToLine)) {
+        buff.push(`${pc}:${line}`);
+      }
```
michaeldiamant, line 5355: @barnjamin Optional:  The following statements _can_ be written more concisely using higher order functions.

I don't feel strongly.  Primarily, I tested the refactoring locally to further my understanding of TypeScript language features.  If you prefer as is, feel welcomed to resolve.

```suggestion
      const buff = Object.entries(this.sourcemap.pcToLine).map(([pc, line]) => `${pc}:${line}`);
```

PR #593, opened 2022-06-22
Properly set maxWidth in trace

### LINKED ISSUE #592: Max width is not properly during dry run transaction trace
### Subject of the issue

`maxValueWidth` of the StackPrinterConfig doesn't seem to work. 

### Your environment
algosdk version `1.18.0`

### Steps to reproduce

1. Call `console.log(txn.appTrace({maxValueWidth: 5000, topOfStackFirst: false}))`

### Expected behaviour
The max width of values would be increased from the default width

### Actual behaviour
Values are still truncated with the default width:

```
pc# |ln# |source                            |scratch |stack
1   |1   |pushbytes 0x6d657373616765 // ... |        |[]
10  |2   |pushbytes 0x68656c6c6f20776f72... |        |[0x6d657373616765]
24  |3   |app_global_put                    |        |[0x6d657373616765, 0x68656c6c6...
25  |4   |pushbytes 0x746869732069732061... |        |[]
59  |5   |log                               |        |[0x7468697320697320616e20696d7...
60  |6   |pushint 1                         |        |[]
62  |7   |return                            |        |[1]
63  |8   |                                  |        |[1]
```

As a sanity check I made sure that the spc object was getting passed correctly by changing `topOfStackFirst` which does indeed work. It seems to be a problem specific to the `maxValueWidth`

joe-p:
This fixes #592. The `maxWidth` was simply never updated if a non-default value was provided in the spc. This PR ensures that `maxWidth` is properly set. Tested with default `maxWidth` and a specific `spc.maxValueWidth` and functionality now works as expected.

### DIFF
--- a/src/dryrun.ts
+++ b/src/dryrun.ts
@@ -365,8 +365,7 @@ class DryrunTransactionResult {
     disassembly: string[],
     spc: StackPrinterConfig
   ): string {
-    let maxWidth = defaultMaxWidth;
-    if (spc.maxValueWidth === undefined) maxWidth = spc.maxValueWidth;
+    const maxWidth = spc.maxValueWidth || defaultMaxWidth;
 
     // Create the array of arrays, each sub array contains N columns
     const lines = [['pc#', 'ln#', 'source', 'scratch', 'stack']];


algochoi (approved):
thanks for the fix!

PR #595, opened 2022-06-23
API: Support attaching signatures to standard and multisig transactions

jdtzmn:
This PR adds an `attachSignature` method to the Transaction class and related external signature methods for multisig transactions.

### DIFF
--- a/src/multisig.ts
+++ b/src/multisig.ts
@@ -42,8 +42,8 @@ interface MultisigMetadataWithPks extends Omit<MultisigMetadata, 'addrs'> {
 }
 
 /**
- * createMultisigTransaction creates a multisig transaction blob.
- * @param txnForEncoding - the actual transaction to sign.
+ * createMultisigTransactionWithSignature creates a multisig transaction blob with an included signature.
+ * @param txn - the actual transaction to sign.
  * @param rawSig - a Buffer raw signature of that transaction
  * @param myPk - a public key that corresponds with rawSig
  * @param version - multisig version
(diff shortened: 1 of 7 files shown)


src/transaction.ts
```diff
@@ -1035,6 +1035,19 @@ export class Transaction implements TransactionStorageStructure {
     return new Uint8Array(encoding.encode(sTxn));
   }
 
+  attachSignature(signerAddr: string, signature: Uint8Array) {
+    const sTxn: EncodedSignedTransaction = {
```
jasonpaulos, line 1039: Perhaps verify the length of the signature? I believe it should always be 64 bytes

src/multisig.ts
```diff
@@ -391,6 +411,40 @@ export function appendSignMultisigTransaction(
...
+ * This makes it possible to compile a multisig signature using only raw signatures from external methods.
+ * @param multisigTxnBlob - an encoded multisig txn. Supports non-payment txn types.
+ * @param version - multisig version
+ * @param threshold - multisig threshold
+ * @param addrs - a list of Algorand addresses representing possible signers for this multisig. Order is important.
+ * @param signerAddr - address of the signer
+ * @param signature - raw multisig signature
+ * @returns object containing txID, and blob representing encoded multisig txn
+ */
+export function appendSignRawMultisigSignature(
```
jasonpaulos, line 425: This looks good, but we're still missing an alternative to `signMultisigTransaction` to create the `multisigTxnBlob`

src/multisig.ts
```diff
@@ -147,6 +147,26 @@ export class MultisigTransaction extends txnBuilder.Transaction {
...
+   * @param signerAddr - address of the signer
+   * @param signature - raw multisig signature
+   * @returns an encoded, partially signed multisig transaction.
+   */
+  partialSignWithMultisigSignature(
+    metadata: MultisigMetadataWithPks,
+    signerAddr: string,
+    signature: Uint8Array
+  ) {
+    return createMultisigTransaction(
```
jasonpaulos, line 163: Perhaps verify signature length here too?

jasonpaulos (commented):
Nearly there, I just have a few suggestions

jasonpaulos (approved):
Looks good!

PR #597, opened 2022-06-24
Add sourcemap query param to compile

MacroChip:
For chaining and convenience


barnjamin:
hey @MacroChip, thanks!

I did include this in https://github.com/algorand/js-algorand-sdk/pull/590 which also includes a class to decode the result into something directly usable

MacroChip:
Sounds good. Cheers

PR #598, opened 2022-06-26
Remove code that relies on node's path module

bmdelacruz:
With the removal of the dependency to node's path module, there's one less thing to 'browserify'. All tests, especially the ones under client url construction, passed. I'll leave the the webpack configuration updates to you all, I guess...

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -33,17 +33,7 @@ With Vite, you would see:
 Uncaught ReferenceError: Buffer is not defined
 ```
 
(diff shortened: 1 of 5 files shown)


jasonpaulos (commented):
Hi @bmdelacruz, thank you for working on this. This seems like a welcome addition to the SDK.

My only comment is, could you also remove the references to `path-browserify` in https://github.com/algorand/js-algorand-sdk/blob/develop/FAQ.md#vite-projects ?

jasonpaulos (approved):
Thank you for the changes

PR #599, opened 2022-06-26
fix: safe intDecoding

AlgoDoggo:
This PR fixes a couple of bugs with safe mode for integer decoding during json parse:
- safe mode threw errors if `10**15 > value < Number.MAX_SAFE_INTEGER `. It would have shown in the unit test if we had tested for edge cases.
- safe mode did not decode integers as numbers. It didn't alter the output of JSONbig in any way.
It is again only an issue for those values that are `10**15 > value < Number.MAX_SAFE_INTEGER `


jasonpaulos (approved):
Thank you for finding and fixing this bug

PR #601, opened 2022-06-30
perf: waitForConfirmation

AlgoDoggo:
This PR addresses some undesirable behavior in the waitForConfirmation function.
- Separation of concerns. The function as it was would throw an error when unable to fetch the node status. This behavior is undocumented and dApps may consider that an error from waitForConfirmation means the transaction has been rejected, when it could be an unrelated issue.
-  It is possible to initialize the loop with a currentRound of 0, fixing the above issue and saving an api call in the process. This is what I did here.

### DIFF
--- a/src/wait.ts
+++ b/src/wait.ts
@@ -21,7 +21,7 @@ export async function waitForConfirmation(
   let counter = 0;
 
   /* eslint-disable no-await-in-loop */
-  while (counter < waitRounds) {
+  while (counter <= waitRounds) {
     let pendingInfo = {};
     try {
       pendingInfo = await client.pendingTransactionInformation(txid).do();
@@ -42,7 +42,7 @@ export async function waitForConfirmation(
     }
 
     const status = await client.statusAfterBlock(currentRound).do();
-    currentRound = status['last-round'] + 1;
+    currentRound = ((status && status['last-round']) || currentRound) + 1;
     counter += 1;
   }
   /* eslint-enable no-await-in-loop */
--- a/src/wait.ts
+++ b/src/wait.ts
@@ -42,6 +42,10 @@ export async function waitForConfirmation(
     }
 
     const status = await client.statusAfterBlock(currentRound).do();
+    // wait 2 seconds if we do not get any status back or if currentRound is 0
+    if (!status || !currentRound) {
+      await new Promise<void>((res) => setTimeout(res, 2000));
+    }
     currentRound = ((status && status['last-round']) || currentRound) + 1;
     counter += 1;
   }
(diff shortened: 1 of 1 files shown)


src/wait.ts
```diff
@@ -17,41 +17,33 @@ export async function waitForConfirmation(
...
+      // Got the completed Transaction
+      return pendingInfo;
+    }
+
+    if (pendingInfo['pool-error']) {
+      // If there was a pool error, then the transaction has been rejected
+      throw new Error(`Transaction Rejected: ${pendingInfo['pool-error']}`);
+    }
+
+    const status = await client.statusAfterBlock(currentRound).do();
```
algochoi, line 44: It feels prudent to throw the old error here, i.e.

```
if (typeof status === 'undefined') {
    throw new Error('Unable to get node status');
}
```

algochoi (commented):
Thanks for the PR! The changes look okay to me. 

I feel like it's a slight improvement, but it looks like we'd have to do an extra API call if we are not able to get the node status (and throw an error) anyway. I'll defer to @jasonpaulos @barnjamin  if they have any strong opinions on it.

src/wait.ts
AlgoDoggo: So the idea of the PR is precisely not to throw an error if there is a temporary network issue before or during an iteration of the loop. If status is undefined the current round will be incremented by 1 in the below line and the loop will have a chance to keep going, the only downside being that the next iteration will happen sooner that it otherwise would have.

jdtzmn: I can see why it's valuable to not throw an error if there is a temporary network issue, but wouldn't it then be beneficial to include a timeout before the next while loop iteration? Otherwise, as you noted, you're giving no time for the network issue to resolve itself.

AlgoDoggo: @jdtzmn Good idea, I've also added a timeout for the first iteration of the loop since `statusAfterBlock` will return almost instantly when `currentRound` is 0.
Could you set the appropriate label for the ci test to pass pls?

jasonpaulos:
Hi @AlgoDoggo, thank you for making this PR and helping us understand the issue.

> * Separation of concerns. The function as it was would throw an error when unable to fetch the node status. This behavior is undocumented and dApps may consider that an error from waitForConfirmation means the transaction has been rejected, when it could be an unrelated issue.

I definitely agree this is a problem that needs to be addressed. However, after internal discussion, we believe a more appropriate solution would be to better allow the caller to determine why a call to `waitForConfirmation` fails, rather than eliminating all possible failures except for a pending transaction rejection.

Our reasoning is that if `waitForConfirmation` must continue even in the presence of network failures or algod status issues, its complexity grows and becomes more difficult for callers to reason about and for us to maintain. And even if this is the route we chose, it's still possible for callers to confuse an actual transaction rejection with the timeout error.

Unfortunately, because we've decided on a different type of solution, I'm going to close this PR. In its place, I've created issue #615 which describes in more detail the problem and type of solution we're looking for.

PR #604, opened 2022-07-06
Boxes: Add support for Boxes

algochoi:
Adds SDK support for Boxes introduced in https://github.com/algorand/indexer/pull/1168 and https://github.com/algorand/go-algorand/pull/4149.


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="feature/box-storage"
```
ahangsu, line 3: TODO: when algorand-sdk-testing PR merged:
```suggestion
SDK_TESTING_BRANCH="master"
```

jasonpaulos (commented):
I plan to make a child PR to resolve the issues I pointed out here

src/boxStorage.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * translateBoxReferences translates an array of BoxReferences with app IDs
+ * into an array of EncodedBoxReferences with foreign indices.
+ */
+export function translateBoxReferences(
+  references: BoxReference[],
+  foreignApps: number[],
+  appIndex: number
+): EncodedBoxReference[] {
+  if (!Array.isArray(references)) return [];
```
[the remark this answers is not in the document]
algochoi: `references` should be an array of `BoxReference`, but I added this in case a `null` or `undefined` is passed in and to protect the `.map()` method from failing (alternatively we can just check for `null`?).

src/types/transactions/base.ts
```diff
@@ -130,6 +130,21 @@ export interface SuggestedParams {
...
 }
 
+/**
+ * A grouping of the app ID and name of the box in an Uint8Array
+ */
+export interface BoxReference {
+  /**
+   * A unique application index
+   */
+  appIndex: number;
```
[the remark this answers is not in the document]
jasonpaulos: I'm going to walk this back, as our current transaction class design makes union types difficult to deal with. In the future when we consider changing the other instances of `number` app/asset types, this should be included in the discussion.

src/makeTxn.ts
```diff
@@ -1119,6 +1120,7 @@ export function makeApplicationCreateTxn(
   accounts?: AppCreateTxn['appAccounts'],
   foreignApps?: AppCreateTxn['appForeignApps'],
   foreignAssets?: AppCreateTxn['appForeignAssets'],
+  boxes?: AppCreateTxn['boxes'],
```
jasonpaulos, line 1123: Let's avoid a breaking change by moving this argument to the end.

src/makeTxn.ts
```diff
@@ -1241,6 +1247,7 @@ export function makeApplicationUpdateTxn(
   accounts?: AppUpdateTxn['appAccounts'],
   foreignApps?: AppUpdateTxn['appForeignApps'],
   foreignAssets?: AppUpdateTxn['appForeignAssets'],
+  boxes?: AppUpdateTxn['boxes'],
```
jasonpaulos, line 1250: Same here

src/makeTxn.ts
```diff
@@ -1338,6 +1349,7 @@ export function makeApplicationDeleteTxn(
   accounts?: AppDeleteTxn['appAccounts'],
   foreignApps?: AppDeleteTxn['appForeignApps'],
   foreignAssets?: AppDeleteTxn['appForeignAssets'],
+  boxes?: AppDeleteTxn['boxes'],
```
jasonpaulos, line 1352: And here

src/makeTxn.ts
```diff
@@ -1427,6 +1443,7 @@ export function makeApplicationOptInTxn(
   accounts?: AppOptInTxn['appAccounts'],
   foreignApps?: AppOptInTxn['appForeignApps'],
   foreignAssets?: AppOptInTxn['appForeignAssets'],
+  boxes?: AppOptInTxn['boxes'],
```
jasonpaulos, line 1446: And here

src/makeTxn.ts
```diff
@@ -1516,6 +1537,7 @@ export function makeApplicationCloseOutTxn(
   accounts?: AppCloseOutTxn['appAccounts'],
   foreignApps?: AppCloseOutTxn['appForeignApps'],
   foreignAssets?: AppCloseOutTxn['appForeignAssets'],
+  boxes?: AppCloseOutTxn['boxes'],
```
jasonpaulos, line 1540: And here

src/makeTxn.ts
```diff
@@ -1605,6 +1631,7 @@ export function makeApplicationClearStateTxn(
   accounts?: AppClearStateTxn['appAccounts'],
   foreignApps?: AppClearStateTxn['appForeignApps'],
   foreignAssets?: AppClearStateTxn['appForeignAssets'],
+  boxes?: AppClearStateTxn['boxes'],
```
jasonpaulos, line 1634: And here

src/makeTxn.ts
```diff
@@ -1694,6 +1725,7 @@ export function makeApplicationNoOpTxn(
   accounts?: AppNoOpTxn['appAccounts'],
   foreignApps?: AppNoOpTxn['appForeignApps'],
   foreignAssets?: AppNoOpTxn['appForeignAssets'],
+  boxes?: AppNoOpTxn['boxes'],
```
jasonpaulos, line 1728: And here

src/makeTxn.ts
jasonpaulos: Resolved by #660

src/client/v2/indexer/searchForApplicationBoxes.ts
```diff
@@ -0,0 +1,134 @@
+import JSONRequest from '../jsonrequest';
+import HTTPClient from '../../client';
+import IntDecoding from '../../../types/intDecoding';
+import { BoxDescriptor } from '../algod/models/types';
+
+export interface SearchForApplicationBoxesResponse {
```
jasonpaulos, line 6: I'm leaving this comment as a visual reminder to look into indexer response models more.

Right now this SDK only uses algod's REST API definitions to generate the models, which means some indexer-specific models like this one are missing. Ideally we would be able to generate these models as well instead of manually creating them here.

jasonpaulos: I opened #662 which addresses this

jasonpaulos (approved):
I contributed to this PR, so my approval should not be taken by itself

.test-env
michaeldiamant, line 3:
```suggestion
SDK_TESTING_BRANCH="master"
```

PR #610, opened 2022-07-16
docs: tealSign

AlgoDoggo:
I'm using `ed25519verify` in a stateful app and it took a while to understand the third argument to provide `tealSign` is not the contract address but the program hash. Those are equals for LogicSigs but not for apps. I updated the doc to make it clear.


AlgoDoggo:
We don't have permissions to set labels for PRs, so they auto-fail ci tests.

AlgoDoggo:
@barnjamin would you add a label pls?

barnjamin (approved):

algochoi (approved):

PR #614, opened 2022-07-22
Testing: Use Dev mode network for cucumber tests

algochoi:
This PR aims to allow Dev mode network to be used in cucumber tests to speed up testing time. 

Cucumber's `this` and `World` objects seem to be bound to the steps, so I refrained from using helper methods that rely on state associated with the cucumber's world/this object. This PR also removes some `statusAfterBlock` v1 API calls, as they seem to be more flaky in dev mode, and often a redundant step in the scheme of the entire cucumber scenario.

PR #616, opened 2022-07-29
Remove unnecessary use of v2 algod client

michaeldiamant:
Extends #614 to remove what appears to be unnecessary v2 algod client usage.  The initialization of a v2 client in _an algod client_ led me to creating the PR.

Note - Predating the PR, there is a flaky test:  _Scenario: Getting transaction by ID_.  The PR does not attempt to remedy it.

### DIFF
--- a/tests/cucumber/steps/steps.js
+++ b/tests/cucumber/steps/steps.js
@@ -881,15 +881,17 @@ module.exports = function getSteps(options) {
   });
 
   Then('the transaction should go through', async function () {
-    waitForAlgodInDevMode();
-    const info = await this.acl.pendingTransactionInformation(this.txid);
+    let info = await this.acl.pendingTransactionInformation(this.txid);
     assert.deepStrictEqual(true, 'type' in info);
-    // info = await this.acl.transactionById(this.txid);
-    // assert.deepStrictEqual(true, 'type' in info);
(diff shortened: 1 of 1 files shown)


jasonpaulos:
I don't have much context on this PR, but why would we remove v2 algod client usage in favor of v1 (which should be deprecated if it's not already)?

michaeldiamant:
@jasonpaulos Responding inline. If it doesn't address the question, let's speak live:

> why would we remove v2 algod client usage in favor of v1 

* The test in question _is_ a v1 API test.  For reasons unclear to me (that's why I opened the PR), v2 algod client usage is intermingled.  Like I mention in the description, it's strange to see added v2 initialization. 
* If I had to guess, I _think_ v2 client was introduced for convenience of marshalling suggested params.
* When we _finally_ remove v1 API support, I expect we'll make test changes to use v2 client. But, I think it's out of scope for the dev mode network testing effort.

michaeldiamant:
For reviewers:  The build failure associated with https://github.com/algorand/js-algorand-sdk/pull/616/commits/3a8cbb162a242a5fbb34c7ad6309fcfed858a58c should be fixed in #617.

Let's bring #617 in and then the build ought to pass here.

algochoi (approved):
LGTM

PR #617, opened 2022-07-29
Add missing await for async function calls

michaeldiamant:
Extends #614 to add `await` to `waitForAlgodInDevMode` invocations.  Webstorm IDE warnings pointed out the async call is missing `await`.

Additionally, the PR reverts interactions in _the transaction should go through_ to mirror prior to #614.  I think the shuffling in #614 is not necessary.


algochoi (approved):
thanks!

PR #625, opened 2022-08-18
Break out cucumber tags into their own files

tzaffi:
Breaking out tags under
* `tests/cucumber/unit.tags`
* `tests/cucumber/integration.tags`

for more human readable diffs and also to conform with the other SDK's


tests/cucumber/integration.tags
```diff
@@ -0,0 +1,14 @@
...
+@algod
+@applications
+@applications.verified
+@assets
+@auction
+@c2c
+@compile
+@compile.sourcemap
+@dryrun
+@kmd
```
tzaffi, line 11: 3 `@indexer*` integration tags have also been removed

michaeldiamant (approved):
@tzaffi Thanks for bringing the tag separation to js-algorand-sdk.  

Merging to resolve parent PR thread.

PR #629, opened 2022-08-22
enhancement: Initial stateproofs support

Eric-Warehime:
Initial support for state proofs APIs in the JS Client.

`src/client/v2/algod/models/types.ts` was generated using the algorand/generator package.


tests/cucumber/unit.tags
```diff
@@ -16,7 +16,9 @@
...
 @unit.responses.unlimited_assets
 @unit.sourcemap
+@unit.stateproof.paths
+@unit.stateproof.responses
 @unit.tealsign
 @unit.transactions
 @unit.transactions.keyreg
-@unit.transactions.payment
\ No newline at end of file
+@unit.transactions.payment
```
tzaffi, line 24: I'm glad this works!

tests/cucumber/unit.tags
```diff
@@ -16,7 +16,9 @@
 @unit.responses.231
 @unit.responses.unlimited_assets
 @unit.sourcemap
+@unit.stateproof.paths
```
tzaffi, line 19: Will a follow up address the other stateproof tags:  `@unit.stateproof.responses` and `` @unit.stateproof.responses.msgp` ? Or is that not slated to be tested in JS?

src/client/v2/algod/algod.ts
```diff
@@ -25,6 +25,8 @@ import {
   AlgodTokenHeader,
   CustomTokenHeader,
 } from '../../urlTokenBaseHTTPClient';
+import LightBlockHeaderProof from './lightBlockHeaderProof';
+import StateProof from './stateproof';
```
tzaffi, line 29: not sure about the conventions of this repo, but should these single imports be inserted alphabetically above?

src/client/v2/indexer/lookupAssetTransactions.ts
```diff
@@ -61,7 +61,7 @@ export default class LookupAssetTransactions extends JSONRequest {
    *        .do();
    * ```
    *
-   * @param type - one of `pay`, `keyreg`, `acfg`, `axfer`, `afrz`, `appl`
+   * @param type - one of `pay`, `keyreg`, `acfg`, `axfer`, `afrz`, `appl`, `stpf`
```
tzaffi, line 64: I'm surprised by the need to support filtering asset related transactions by state proofs, but I don't see the big picture either.

tzaffi (commented):
The structure of the PR looks good. I have neither JS nor StateProof expertise, but I did bring up a couple questions.

src/client/v2/indexer/lookupAssetTransactions.ts
```diff
@@ -61,7 +61,7 @@ export default class LookupAssetTransactions extends JSONRequest {
    *        .do();
    * ```
    *
-   * @param type - one of `pay`, `keyreg`, `acfg`, `axfer`, `afrz`, `appl`, `stpf`
```
tzaffi: I should say I'm not confident about my previous question. I could see an argument for keeping this codec. Just wanted to provoke a thought or two.

Eric-Warehime: I don't think that a state proof txn would even have an associated asset, so in that sense lookupAssetTransactions would never return a state proof in the response and so the filter would be a no-op.

That's my understanding of it currently.

src/client/v2/algod/algod.ts
Eric-Warehime: They do look to be sorted somewhat, so I updated that.

tests/cucumber/unit.tags
Eric-Warehime: Added in recent commits.

src/client/v2/algod/algod.ts
```diff
@@ -462,4 +464,36 @@ export default class AlgodClient extends ServiceClient {
   getProof(round: number, txID: string) {
     return new Proof(this.c, this.intDecoding, round, txID);
   }
+
+  /**
+   * Returns a light block header proof for a given block.
```
shiqizng, line 469: I believe descriptions in the java sdk are correct, https://github.com/algorand/java-algorand-sdk/pull/360/files#diff-5b49d198d3ad575154467c557163c27eeab6cd462df1b09fb6d182b1e2f95193

shiqizng (commented):
I don't see the endpoint `/v2/blocks/{round}/transactions/{txid}/proof` added in the PR

Eric-Warehime:
> I don't see the endpoint `/v2/blocks/{round}/transactions/{txid}/proof` added in the PR

It was added previously https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/v2/algod/proof.ts#L19

tests/cucumber/unit.tags
Eric-Warehime: Decided to remove these--I got them to work in browserless mode, but the firefox tests end up failing.

I did a bunch of digging into this and could not end up getting msgp decoding working for them in the returned responses within the firefox environment.

tzaffi: @jasonpaulos - I believe you have a deeper understanding of the browser-based tests so I wanted to ping you into the discussion as well.

tzaffi (approved):
All my questions have been addressed but I recommend getting approval from people more familiar with state proofs as well.

src/client/v2/algod/algod.ts
```diff
@@ -462,4 +464,36 @@ export default class AlgodClient extends ServiceClient {
   getProof(round: number, txID: string) {
```
algoidan, line 464: This endpoint now has one more optional parameter, I think we should reflect that as well.
In addition, even though the rest endpoint url itself (`/v2/blocks/${this.round}/transactions/${this.txID}/proof`) stays the same  I think it is better to rename it to `getTransactionProof` or similar since it the name doesn't suggest what proof you get.

src/client/v2/algod/algod.ts
```diff
@@ -462,4 +464,36 @@ export default class AlgodClient extends ServiceClient {
...
+   *
+   * [Response data schema details](https://developer.algorand.org/docs/rest-apis/algod/v2#get-v2blocksroundlightheaderproof)
+   * @param round
+   */
+  getLightBlockHeaderProof(round: number) {
+    return new LightBlockHeaderProof(this.c, this.intDecoding, round);
+  }
+
+  /**
+   * Returns a state proof for a given block.
```
algoidan, line 485:
```suggestion
   * Gets a state proof that covers a given round
```

algoidan (commented):
Overall looks pretty good.
Need to keep in mind that state proofs introduce a new transaction. Currently, the msgpack decoder would fail if it comes across a state proof transaction. Think we should add basic support for the state proof transaction - load the state proof message and the proof itself into a byte array (similar to what we do on the python sdk)

Eric-Warehime:
> Overall looks pretty good. Need to keep in mind that state proofs introduce a new transaction. Currently, the msgpack decoder would fail if it comes across a state proof transaction. Think we should add basic support for the state proof transaction - load the state proof message and the proof itself into a byte array (similar to what we do on the python sdk)

Added message, type, and proof to the txn. The python leaves out the added block header fields, so I've done the same here.

src/transaction.ts
```diff
@@ -198,6 +201,9 @@ export class Transaction implements TransactionStorageStructure {
   nonParticipation?: boolean;
   group?: Buffer;
   extraPages?: number;
+  stateProofType?: number | bigint;
```
algoidan, line 204: why not use just `number`?

src/transaction.ts
```diff
@@ -826,6 +853,42 @@ export class Transaction implements TransactionStorageStructure {
...
+      if (!txn.apid) delete txn.apid;
+      if (!txn.apaa || !txn.apaa.length) delete txn.apaa;
+      if (!txn.apap) delete txn.apap;
+      if (!txn.apsu) delete txn.apsu;
+      if (!txn.apan) delete txn.apan;
+      if (!txn.apfa || !txn.apfa.length) delete txn.apfa;
+      if (!txn.apas || !txn.apas.length) delete txn.apas;
+      if (!txn.apat || !txn.apat.length) delete txn.apat;
+      if (!txn.apep) delete txn.apep;
+      if (txn.grp === undefined) delete txn.grp;
```
algoidan, line 889: don't we want to remove the `grp` if it exists?  (i.e `if !(txn.grp === undefined)`)

algoidan:
looking at the code, it looks like the `js` code does define a block header structure (https://github.com/algorand/js-algorand-sdk/blob/develop/src/types/blockHeader.ts)
should we add the state-proof tracking + the txn256?

src/transaction.ts
Eric-Warehime: We use the `number | bigint` union type for `assetTotal` and `amount` which are also `uint64` underlying types. Seems fine to use it here as well.

src/transaction.ts
Eric-Warehime: The encoded transaction shouldn't include zero or empty values, so this is just removing fields which were potentially initialized, but have no values.

Eric-Warehime:
> looking at the code, it looks like the `js` code does define a block header structure (https://github.com/algorand/js-algorand-sdk/blob/develop/src/types/blockHeader.ts) should we add the `state-proof tracking` + `txn256` to it?

Added.

tests/cucumber/unit.tags
jasonpaulos: Resolved in afbd042

algoidan (approved):

PR #632, opened 2022-08-24
Enhancement: Deprecating use of langspec

ahangsu:


.test-env
```diff
@@ -1,6 +1,7 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+# TODO change back to master when sdk-testing is merged
+SDK_TESTING_BRANCH="deprecate-langspec"
```
ahangsu, line 4: after algorand-sdk-testing PR is merged, do:
```suggestion
SDK_TESTING_BRANCH="master"
```

src/logic/logic.ts
```diff
@@ -149,8 +160,48 @@ function readPushByteOp(
   return [size, byteArray];
 }
 
+/** sanityCheckProgram performs heuristic program validation:
+ * check if passed in bytes are Algorand address or is B64 encoded, rather than Teal bytes
+ *
+ * @param program - Program bytes to check
+ * @throws
+ * @returns
```
jasonpaulos, line 168: nit: `@throws` and `@returns` should be removed if you don't wish to add explanation to them.

src/logic/logic.ts
```diff
@@ -149,8 +160,48 @@ function readPushByteOp(
...
+  const blankSpaceOrd = ' '.charCodeAt(0);
+  const tildeOrd = '~'.charCodeAt(0);
+
+  const isPrintable = (x: number) => blankSpaceOrd <= x && x <= tildeOrd;
+  const isAsciiPrintable = program.every(
+    (x: number) => x === lineBreakOrd || isPrintable(x)
+  );
+
+  if (isAsciiPrintable) {
+    const programStr = Buffer.from(program.buffer).toString();
```
jasonpaulos, line 183: You should use `Buffer.from(program)` instead of `Buffer.from(program.buffer)`. It's possible that the `Uint8Array` `program` represents only a substring of the underlying `.buffer` object.

tests/cucumber/steps/steps.js
```diff
@@ -4454,6 +4454,33 @@ module.exports = function getSteps(options) {
...
+      this.actualErrMsg = e.message;
+    }
+  });
+
+  Then(
+    'if the heuristic sanity check throws an error, the error contains {string}',
+    async function (errMsg) {
+      if (errMsg !== '')
+        assert.strictEqual(this.actualErrMsg.includes(errMsg), true);
+      else assert.strictEqual(this.actualErrMsg === undefined, true);
```
jasonpaulos, line 4480: nit: you can use `assert.ok(this.actualErrMsg.includes(errMsg))` and `assert.strictEqual(this.actualErrMsg, undefined)`, which are a little more concise

src/logic/logic.ts
ahangsu: good point, removing

src/logic/logic.ts
ahangsu: this actually blows my mind, ok I will just pass in `program` then.

tests/cucumber/steps/steps.js
ahangsu: good point

michaeldiamant (approved):

jasonpaulos (approved):

PR #656, opened 2022-09-21
Box storage changes

jasonpaulos:
Child PR to #604. This PR contains changes from my review.

Additionally, I modified the return value types for the new algod and indexer box endpoints. This isn't strictly necessary to be consistent with our other endpoints, but this improvement gives the responses types (as supposed to `any`) and base64 decodes any box names and values, which I think will greatly help users.

### DIFF
--- a/src/client/v2/algod/algod.ts
+++ b/src/client/v2/algod/algod.ts
@@ -440,15 +440,21 @@ export default class AlgodClient extends ServiceClient {
    * ```typescript
    * const index = 60553466;
    * const boxName = Buffer.from("foo");
-   * const boxValue = await algodClient.getApplicationBoxByName(index).name(boxName).do();
+   * const boxResponse = await algodClient.getApplicationBoxByName(index, boxName).do();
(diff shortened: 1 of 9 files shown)


src/client/client.ts
```diff
@@ -224,7 +226,8 @@ export default class HTTPClient {
     relativePath: string,
     query?: Query<any>,
     requestHeaders: Record<string, string> = {},
-    jsonOptions: utils.JSONOptions = {}
+    jsonOptions: utils.JSONOptions = {},
+    parseBody: boolean = true
```
algochoi, line 230: Do we want to update the JSDoc for this param?

algochoi (approved):
Looks good, non-blocker comment about the JSDoc

src/client/client.ts
jasonpaulos: Added in e7e928e076a6113574aa8eb53d2e718de96a39df

ahangsu (approved):

Issue #657, opened 2022-09-23
Algod port argument should not have weird default


barnjamin:
I've never seen 4180 as the port for an Algod API. In sandbox its 4001, for APIs it's almost always 443.

If the dev doesn't set it to an empty string, 4180 is used and often results in the call just hanging until timeout with no useful error message.

We should consider setting the default port to an empty string and let the caller override it if they have some special port

Issue #650, opened 2022-09-12
Breaking changes in v2.0.0


michaeldiamant:
Itemizes functionality that will be removed in the upcoming major release.

Task list for algorand-sdk-testing stories:
- [x] https://github.com/algorand/algorand-sdk-testing/issues/201 - https://github.com/algorand/js-algorand-sdk/pull/713
- [x] https://github.com/algorand/algorand-sdk-testing/issues/202 - no need
- [x] https://github.com/algorand/algorand-sdk-testing/issues/223 - https://github.com/algorand/js-algorand-sdk/pull/712
- [x] https://github.com/algorand/algorand-sdk-testing/issues/254 - https://github.com/algorand/js-algorand-sdk/pull/712
- [x] https://github.com/algorand/algorand-sdk-testing/issues/255 - https://github.com/algorand/js-algorand-sdk/pull/714
- [x] https://github.com/algorand/generator/issues/46 - https://github.com/algorand/generator/pull/51
- [x] Regenerate Typescript models via generator - https://github.com/algorand/js-algorand-sdk/pull/720

Issue #458, opened 2021-10-18
algosdk.Algodv2 overwrites empty port parameter


tholcman:
### Subject of the issue

```
const client = new algosdk.Algodv2("DUMMYTOKEN", "http://127.0.0.1:17458");
```

Should create a client which will request port 17458, but the client requests port 4180. 

it is overwritten here:

https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/v2/algod/algod.ts#L27

### Your environment

SDK version 1.12.0, develop branch, ...

### Steps to reproduce

1. Use an URL with defined non-standard port without `port` parameter
2. ...

### Expected behaviour

It should request a algod on "http://127.0.0.1:17458"

### Actual behaviour

it requests "http://127.0.0.1:4180"

winder:
Closing in favor of #657

PR #658, opened 2022-09-23
BugFix: set defaults to empty

barnjamin:
Addresses #657

### DIFF
--- a/src/client/v2/algod/algod.ts
+++ b/src/client/v2/algod/algod.ts
@@ -66,8 +66,8 @@ export default class AlgodClient extends ServiceClient {
       | AlgodTokenHeader
       | CustomTokenHeader
       | BaseHTTPClient,
-    baseServer = 'http://r2.algorand.network',
-    port: string | number = 4180,
+    baseServer: string = '',
+    port: string | number = 0,
     headers: Record<string, string> = {}
   ) {
     super('X-Algo-API-Token', tokenOrBaseClient, baseServer, port, headers);


src/client/v2/algod/algod.ts
```diff
@@ -66,8 +66,8 @@ export default class AlgodClient extends ServiceClient {
       | AlgodTokenHeader
       | CustomTokenHeader
       | BaseHTTPClient,
-    baseServer = 'http://r2.algorand.network',
-    port: string | number = 4180,
+    baseServer: string = '',
```
michaeldiamant, line 69: @barnjamin I agree js-algornad-sdk is inconsistent with other SDKs and is arguably inconvenient.  

My concerns are:
* It's a breaking change.  I need a bit of convincing to understand the _bug_ framing.
* If we're making a breaking change, can we instead mark the `baseServer` and `port` as required parameters?  I think I don't see the case for _reasonable_ default values here.
 
Aside:   We're planning a major version bump for Q4 2022:  https://github.com/algorand/js-algorand-sdk/issues/650.  If we land on categorizing the PR as a breaking change, then we can consider adding it to https://github.com/algorand/js-algorand-sdk/issues/650.

barnjamin: > I need a bit of convincing to understand the bug framing.

I'm not sure this would break anyone but I have no idea why this port was chosen for default. Maybe it'd help to understand that?

I supposed it's not a bug in the sense that it's broken, more a bug in dev expectations.

> If we're making a breaking change, can we instead mark the baseServer and port as required parameters? I think I don't see the case for reasonable default values here.

I think the port should stay optional since most folks wouldn't consider setting 443 on https hosts and empty string or 0 will allow the client to "figure that out"

winder (approved):
Seems reasonable

src/client/v2/algod/algod.ts
winder: This would also address the bug reported by #458

michaeldiamant: Circling back after a verbal with @barnjamin and @jasonpaulos.  Here's where we landed:
* Changing default values in a public method _is_ a breaking change. 
* Let's include an improvement in v2.0.0.  Concretely, @barnjamin will make these changes:
  * Change base from `develop` to `integration/2.0.0` (new branch).  `integration/2.0.0` will be the integration branch until v2.0.0 is released.  We can safely merge into it.
  *  Since we'll make a breaking change, let's improve the status quo by removing `port`.  The user can supply port as needed in `baseServer`. 
* Based on these outcomes, I added https://github.com/algorand/js-algorand-sdk/issues/657 to the v2.0.0 epic (https://github.com/algorand/js-algorand-sdk/issues/650).

barnjamin:
closing in favor of #711

PR #661, opened 2022-09-27
Swap superagent for fetch

PureBrent:
Purpose:
- Replace superagent with the more modern fetch in the urlTokenBaseHTTPClient.ts file. This adheres to the requirement in Chrome Manifest V3 Service Workers which can be found at [https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration) under Service workers. 

Approach:
- Replace packages
        - Add: 		"cross-fetch": "^3.1.5"
        - Remove:	"superagent": "^6.1.0"
- Modify urlTokenBaseHTTPClient.ts
	- Import fetch from 'cross-fetch' and remove superagent import
	- Modify get, post, delete methods to replace request with fetch 
	- Replace superagent formatting methods for response and error to fetch equivalents

### DIFF
--- a/src/client/urlTokenBaseHTTPClient.ts
+++ b/src/client/urlTokenBaseHTTPClient.ts
@@ -1,4 +1,4 @@
-import * as request from 'superagent';
+import fetch from 'cross-fetch';
 import {
   BaseHTTPClient,
   BaseHTTPClientResponse,
@@ -29,7 +29,7 @@ export type TokenHeader =
 
 /**
  * Implementation of BaseHTTPClient that uses a URL and a token
- * and make the REST queries using superagent.
+ * and make the REST queries using fetch.
  * This is the default implementation of BaseHTTPClient.
  */
 export class URLTokenBaseHTTPClient implements BaseHTTPClient {
@@ -79,25 +79,17 @@ export class URLTokenBaseHTTPClient implements BaseHTTPClient {
     return address.toString();
   }
 
-  /**
-   * Convert a superagent response to a valid BaseHTTPClientResponse
-   * Modify the superagent response
-   * @private
-   */
-  private static superagentToHTTPClientResponse(
-    res: request.Response
-  ): BaseHTTPClientResponse {
-    if (res.body instanceof ArrayBuffer) {
-      // Handle the case where the body is an arraybuffer which happens in the browser
-      res.body = new Uint8Array(res.body);
+  private static async formatFetchResponse(res: any): Promise<Uint8Array> {
+    const contentType = res.headers.get('content-type');
+    if (contentType && contentType.indexOf('application/json') !== -1) {
+      return Buffer.from(JSON.stringify(await res.json()));
     }
-    return res;
+
+    // Failures are expected to be message packed so transform it to Uint8Array
+    return new Uint8Array(await res.arrayBuffer());
   }
 
-  /**
-   * Make a superagent error more readable. For more info, see https://github.com/visionmedia/superagent/issues/1074
-   */
-  private static formatSuperagentError(err: any): Error {
+  private static formatFetchError(err: any): Error {
     if (err.response) {
       try {
         const decoded = JSON.parse(Buffer.from(err.response.body).toString());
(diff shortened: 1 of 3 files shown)


src/client/urlTokenBaseHTTPClient.ts
```diff
@@ -29,7 +29,7 @@ export type TokenHeader =
```
michaeldiamant, line 29: @PureBrent Unrelated to the line, but since the PR is marked ready for review, I want to sanity check:  Have you seen that the CI build failed?

Admittedly, I'm still grokking the PR, though the failing tests do _not_ appear spurious to me.  And I thought I'd raise a note to double check if it's on your radar.

PureBrent: The 2 errors that exist in Chrome and Firefox I understand. With the one regarding a 400 != 200 error being because the test is adding a statusCode instead of status which fetch returns, so I can change the test. The other is a small code change. 

However, the failures in the node test are more mysterious. The failing tests are using a mock server, but make calls that have a path and headers as:
  Path:  http://localhost:31337/health
  HEADERS:  { 'X-Indexer-API-Token': '', accept: 'application/json' }

The fetch code responds with the algod health call which has a status 200 and null body, but the superagent code responds and contains a body that has JSON values which change for each call despite them all being to the health endpoint. I assume this is a result of the mock server overriding paths, but am unsure why these do not work with fetch. Any insight?

algochoi: I took a look into this, and I think it's related to this issue: https://github.com/node-fetch/node-fetch/issues/665

It seems like the failed tests have extremely large response bodies ([example test](https://github.com/algorand/algorand-sdk-testing/blob/master/features/unit/responses.feature#L12)), and this hangs `response.json()`. I replicated this on my machine by confirming that: 1) all test cases with small response bodies are parsed correctly in node; 2) If the response body for the failed test cases is shrunk, then the cases eventually pass without timing out.

purestaketdb: For the invalid content-type responses, can these be addressed so workarounds are not necessary? 

"I've seen that many tests failed and put a change in to address those. It was resulting from responses having a "content-type" as "application/json" when they were msgpack responses."

algochoi: It looks like the main problem is that Node has a smaller default stream buffer size compared to the browser versions (16KB vs 1MB), and I think this limit is hit when cloning the response (i.e. calling `Response.clone()`). 

[A solution](https://github.com/node-fetch/node-fetch/blob/main/README.md#custom-highwatermark) is to increase the `highwatermark` value in the request to 1MB. Unfortunately, I took a look at the `cross-fetch` package, and it lists `node-fetch` v2.x [as a dependency](https://github.com/lquixada/cross-fetch/blob/9fb71cfc2fd5de86d541954f4496113dc29a903e/package.json#L69), and custom `highwatermark` values are only settable in `node-fetch` v3.x. I guess one solution might be to use the newer version of `node-fetch` (inline, or otherwise).

> For the invalid content-type responses, can these be addressed so workarounds are not necessary?

I don't think the content-type headers can be trusted in the wild either, but I agree that fixing these headers in the test cases would reduce the response stream, since no cloning would be needed in the current implementation. If we are sure that response sizes are always <16KB, then this might be a viable approach. But I am not sure of this yet.

PureBrent: Thanks @algochoi, this is very similar to what I've seen. The asynchronous promise for both the response and cloned response will fail when content-type is not json, which means the high watermark and updated node-fetch is the solution if the content-type can't be relied upon in the wild and especially if responses will be greater than 16KB. 

The cross-fetch library doesn't appear to have active plans on updating to v3+ of node-fetch so I have been testing working with using node-fetch as a direct dependency. However, it appears that that requires changes in dozens of files in addition to webpack, utils and build settings changes to work since node-fetch v3+ uses ESM. I have some concern that the expected narrow scope changes for fetch appear they require many changes to be able to work.

jasonpaulos: I started a new PR based on this one where I was able to sidestep the `Response.clone()` issue.

Local testing indicates that PR is able to use fetch and pass all tests without error -- we'll see if CI passes as well.

PureBrent: Very nice., thanks for taking a look at it.

jasonpaulos:
I'm going to close this PR since #676 was just merged. @PureBrent thanks again for bringing this to our attention and contributing!

PR #663, opened 2022-09-29
Enhancement: Code generation improvements

jasonpaulos:
This PR contains improvements from https://github.com/algorand/generator/pull/42.


algochoi (approved):
Looks good - is our new approach to always use generated models for algod/indexer responses going forward?

jasonpaulos:
Generally speaking, yes, I think we should add a `prepare` method to any new request classes which uses the generated models to decode the response. For example, https://github.com/algorand/js-algorand-sdk/blob/c10459092946e1ed8c7317e76390a33d99a3024d/src/client/v2/algod/getApplicationBoxByName.ts#L45-L48

michaeldiamant (approved):
I visually spot-checked several generated examples confirming they match the OpenAPI specification + existing parts of the repo.

ahangsu (approved):

PR #675, opened 2022-10-21
fix: mergeMultisigTransactions logic error

AlgoDoggo:
This bug will manifest when the length of input array is > 2. Because all the mocha tests try to merge only two transactions it wasn't caught.

Essentially you have a for loop that iterates over all the transactions and at every iteration the `newSubsigs` array which should comprise all the pk and signatures from all the txs is overwritten with the the one from the reference case which is `multisigTxnBlobs[0]` + the ones from the current iteration of the loop. Thus when exiting the for loop the `newSubsigs` array is made of the signatures available in the reference tx + the last tx. All the other transactions signatures get overwritten.

It it is possible to fix this bug with a one-liner, right at the beginning of the map it would be enough to replace `current` with
```js
const current = newSubsigs[index];
```

However this whole function in general is a little clunky so I couldn't help refactoring slightly the `map` statement to avoid nested if statement and such. The for loop too can be instantiated at index 1 since index 0 is the base case calculated above. If you don't like style changes the one-liner above is enough to do the job.


AlgoDoggo:
@barnjamin could I have a bug-fix label so cicd passes plz?

jasonpaulos:
Hi @AlgoDoggo, thanks for making this PR. I only have one request: can you add a new test that merges > 2 multisigs, since you said that was missing from our test suite?

I need to take a closer look at the code changes to be absolutely sure they work as intended, and a new test would help with that effort (in addition to being a regression test).

scholtz:
Please fix mergeMultisigTransactions .. it took us also quite some time that documented functionality does not work.

AlgoDoggo:
Hey, I've refactored a test to merge three tx instead of two. You can run it solo against the old and new logic to see the diff.

src/multisig.ts
```diff
@@ -248,8 +248,8 @@ export function mergeMultisigTransactions(multisigTxnBlobs: Uint8Array[]) {
     address.fromMultisigPreImg(refPreImage)
   );
 
-  let newSubsigs = refSigTx.msig.subsig;
-  for (let i = 0; i < multisigTxnBlobs.length; i++) {
+  const newSubsigs = refSigTx.msig.subsig;
```
jasonpaulos, line 251: Since this new code modifies `newSubsigs` directly, I'd feel better if we took a deep copy of `refSigTx.msig.subsig` instead of just an alias. This should do it:
```suggestion
  const newSubsigs: EncodedSubsig[] = refSigTx.msig.subsig.map(({ pk, s }) => ({
    pk,
    s,
  }));
```

src/multisig.ts
```diff
@@ -281,32 +281,21 @@ export function mergeMultisigTransactions(multisigTxnBlobs: Uint8Array[]) {
...
+        throw new Error(MULTISIG_MERGE_SIG_MISMATCH_ERROR_MSG);
       }
-      if (uniSubsig.s) {
-        return {
-          pk: current.pk,
-          s: uniSubsig.s,
-        };
-      }
-      return current;
+      current.s = uniSubsig.s;
```
jasonpaulos, line 298: Minor, as this doesn't affect correctness: this line always modifies `current`, but really it only needs to modify it if `current.s` is undefined.

jasonpaulos (commented):
I believe this code change is correct, and I have independently verified the encoded test transactions.

I only have two small comments about the maintainability of the code.

src/multisig.ts
AlgoDoggo: Do you feel strongly about this? map will do a shallow copy of the array, and destructuring pk and s and returning them in a new object won't create new Uint8Arrays for their values, they will still point to the same reference. For a deep copy I'd need to instantiate new Uint8Arrays , add a check to see if s is undefined etc. It adds clutter + takes some extra memory.

src/multisig.ts
AlgoDoggo: Same question as above, this is a O(1) operation that is sometime redundant, I could add an if statement that would run... in constant time. 
This is a little against my coding style as I try to write things the most concise way but I can make the change if you feel it adds clarity.

AlgoDoggo:
At some point we'll need to run typescript in strict mode, or at least with strictNullChecks enabled. A lot of values that can be undefined are just assumed to exist now by the sdk. It's bound to cause runtime errors eventually.

src/multisig.ts
jasonpaulos: Since we never modify the Uint8Arrays I'm ok with them not being copied.

I think this suggestion, while not strictly necessary, will reduce the likelihood of future mistakes, since it's not obvious that `refSigTx.msig.subsig` is being muted along with `newSubsigs`. Alternatively, if you wanted to add a comment above the definition of `newSubsigs` that modifications to it also affect `refSigTx.msig.subsig`, I view that as acceptable also.

src/multisig.ts
jasonpaulos: I just think it's a bit odd that the above lines ensure that `uniSubsig.s` and `current.s` (if defined) are equivalent, then this line assigns them.

However I don't feel strongly about this, so if you prefer to keep it as is, that's fine.

AlgoDoggo: cool ok

AlgoDoggo:
@jasonpaulos Looks like cicd is failing at the `npm ci` step, chromedriver won't install. I'd try rerunning the workflows but I do not have permissions for this.

src/multisig.ts
AlgoDoggo: I kept the spirit of your snippet but didn't commit it as is. By destructuring pk and s and reinjecting them in a new object there was a situation when s didn't exist in the original obj, it would be set as an undefined prop in the new obj, and then to null probably during msgpack encoding at the bottom at the func. This made a bunch of unit-test fails since it wasn't expected. One of those things... 😅

AlgoDoggo:
ah nvm, the dev branch had moved.

src/multisig.ts
jasonpaulos: Good catch, thanks for making that change

jasonpaulos (approved):

PR #677, opened 2022-10-25
API: Fix docs typo

PabloLION:
I found another typo in your docs😂. I tested that it's not overloaded throwing this error `Invalid format for parameter application-id: error binding string parameter: strconv.ParseUint: parsing "undefined": invalid syntax`


jasonpaulos (approved):
Thanks for fixing this

PR #693, opened 2022-11-14
Tests: Migrate v1 algod dependencies to v2 in cucumber tests

algochoi:
Migrates algod v1 tests to use v2.

SDK Testing PR: https://github.com/algorand/algorand-sdk-testing/pull/249

### DIFF
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="remove-v1-tests"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="remove-v1-tests"
+SDK_TESTING_BRANCH="master"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/tests/cucumber/steps/steps.js
+++ b/tests/cucumber/steps/steps.js
@@ -596,11 +596,6 @@ module.exports = function getSteps(options) {
     return this.v2Client.supply().do();
   });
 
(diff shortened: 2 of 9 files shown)


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="remove-v1-tests"
```
algochoi, line 3:
```suggestion
SDK_TESTING_BRANCH="master"
```

tests/cucumber/steps/steps.js
```diff
@@ -426,16 +379,16 @@ module.exports = function getSteps(options) {
...
       const fundingTxnArgs = {
         from: this.accounts[0],
         to: this.rekey,
         amount: DEV_MODE_INITIAL_MICROALGOS,
         fee: sp.fee,
-        firstRound: sp.lastRound + 1,
-        lastRound: sp.lastRound + 1000,
-        genesisHash: sp.genesishashb64,
+        firstRound: sp.firstRound,
+        lastRound: sp.lastRound,
```
ahangsu, line 390: The change looks interesting, looks like this pattern is everywhere in js sdk, that first round used to be `lastRound + 1`, and `lastRound` is `lastRound + 1000`. Is it because of suggested params change?

ahangsu (commented):
The change looks correct to me, and the test is passing, generally looks good. Only one question about first round and last round.

tests/cucumber/steps/steps.js
algochoi: Yeah, I can't comment exactly on the old code, but the v1 `getTransactionParams()` does not natively have a `firstRound` field. Instead, the `suggestParams` in v1 aliases the `firstRound` field to `lastround` and the `lastRound` becomes `lastRound+1000`:
https://github.com/algorand/js-algorand-sdk/blob/6fdb8720633a874662b382b7ed29736f2dce1919/src/client/algod.js#L336

In v2, `firstRound` is there: https://github.com/algorand/js-algorand-sdk/blob/b780daa21485fb059fee76177eb0dfec18cea762/src/types/transactions/base.ts#L100

ahangsu: that is... quite surprising, but the change is making sense to me

ahangsu (approved):

PR #696, opened 2022-11-23
REST API:  Add KV counts to NodeStatusResponse

michaeldiamant:
Adds NodeStatusResponse fields mirroring https://github.com/algorand/go-algorand/pull/4808.


algochoi (approved):
Looks correctly generated, and I can reproduce this locally. LGTM

Issue #497, opened 2022-01-14
ABI decoder does not allow zero-length arrays


jeapostrophe:
https://github.com/algorand/js-algorand-sdk/blob/develop/src/abi/abi_type.ts#L31

but

https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#types --- "\<type\>[\<N\>]: A fixed-length array of length N, where N >= 0. type can be any other type."

`([1-9][\d]*)` should just be `[\d]+`

jasonpaulos:
We can't use `[\d]+` with the code as is, since it would make things like `uint64[0000000001]` appear as a valid type. And we don't want this because method signatures need to have consistent hashes -- the ARC should probably clarify this.

But we could modify the regex in another way to support 0s. Though it's worth noting this type is essentially a noop since you can't store anything in a 0 length static array.

jasonpaulos:
Just made this: https://github.com/fabrice102/ARCs/pull/2

jeapostrophe:
Good point!

https://en.wikipedia.org/wiki/Robustness_principle --- I think you should accept `uint64[00001]` but never generate it =)

jeapostrophe:
This is still not fixed. You should change the regexp:

```
const staticArrayRegexp = /^([a-z\d[\](),]+)\[(0|[1-9][\d]*)]$/;
```

algoanne:
need to look at the other SDKs as well, probably same issue & fix.

jeapostrophe:
Thank you!

ahangsu:
oops github is acting smart here and closed the ticket 😅

PR #698, opened 2022-11-28
Enhancement: allowing zero-length static array

ahangsu:
Fixes https://github.com/algorand/js-algorand-sdk/issues/497

### DIFF
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="zero-length"
+SDK_TESTING_BRANCH="master"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="zero-length"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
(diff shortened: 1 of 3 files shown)


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="zero-length"
```
ahangsu, line 3: TODO: revert before merge into develop
```suggestion
SDK_TESTING_BRANCH="master"
```

michaeldiamant (approved):
@ahangsu Looks correct - Thanks for bringing the fix into js-algorand-sdk!

tests/10.ABI.ts
```diff
@@ -125,7 +125,6 @@ describe('ABI type checking', () => {
       '[][][]',
       'stuff[]',
       // static array
-      'ufixed32x10[0]',
```
algochoi, line 128: I'd just confirm that something like `bool[01]` also fails for the regex.

ahangsu: we only allow `bool[0]` and `bool[1]`, but not `bool[01]`, see comment here: https://github.com/algorand/js-algorand-sdk/issues/497#issuecomment-1013530294

algochoi: Right, I think `bool[01]` should be added here just to check that it correctly throws an error (since it is not allowed).

ahangsu: got it got it, that's a good point, will add testcase to exclude this case

ahangsu: just add new testcase here: [c4d93ec](https://github.com/algorand/js-algorand-sdk/pull/698/commits/c4d93ececf418b579ea9c173054c20e4a5acb2ac)

algochoi (approved):
LGTM!

PR #700, opened 2022-11-29
Bug-Fix: encode ABI string with non-ASCII characters

### LINKED ISSUE #699: Cannot encode ABI string with non-ASCII characters
## Problem

When I try to encode the value `"What’s your favorite tall building?"` as an ABI `string`, an error occurs.

However, `"What's your favorite tall building?"` works fine. Note the difference between `’` and `'` -- the first is not an ASCII character, while the second is, which leads me to believe non-ASCII characters are not supported.

ahangsu:
Fixes #699


algochoi (commented):
Looking good - nit: I'd also test something like "⚽ 🏃" just so it's a bit clearer we are testing for non-ascii characters like emojis

ahangsu:
makes sense, I will add new case once this build passes

algochoi (approved):
nice! :shipit:

michaeldiamant (approved):

jasonpaulos (approved):
Looks great!

PR #701, opened 2022-11-30
ABI:  Refactor ABI encoding test to round-trip

michaeldiamant:
Following #700, the PR replaces discrete encode and decode tests with a single test performing a round-trip encoding-decoding-encoding.  

The intent is to:
* Confirm round-tripping does _not_ regress.  
* Prevent test case divergence.  Previously, encoding and decoding tests exercised independent test cases. 

Additionally, the PR refactors towards _dynamically generated tests_ (https://mochajs.org/#dynamically-generating-tests) so that individual test case failures can be observed.


ahangsu (approved):
lgtm

tests/10.ABI.ts
```diff
@@ -218,214 +219,228 @@ describe('ABI type checking', () => {
...
+      ABIType.from('(bool[],bool[])'),
+      [[], []],
+      new Uint8Array([0, 4, 0, 6, 0, 0, 0, 0])
+    ),
+    newTestCase(
+      ABIType.from('(string,bool,bool,bool,bool,string)'),
+      ['AB', true, false, true, false, 'DE'],
+      new Uint8Array([0, 5, 160, 0, 9, 0, 2, 65, 66, 0, 2, 68, 69])
+    ),
+  ].forEach((testCase) => {
```
algochoi, line 392: Could we add a testcase to test a number inside a tuple, e.g. `(uint8)`.

tests/10.ABI.ts
```diff
@@ -218,214 +219,228 @@ describe('ABI type checking', () => {
...
+      function decodeReturnsBigInt(t: ABIType): boolean {
+        if (t instanceof ABIUintType || t instanceof ABIUfixedType) {
+          return true;
+        }
+        if (t instanceof ABITupleType) {
+          (t as ABITupleType).childTypes
+            .map(decodeReturnsBigInt)
+            .includes(true);
+        } else if (t instanceof ABIArrayStaticType) {
+          return decodeReturnsBigInt((t as ABIArrayStaticType).childType);
```
algochoi, line 422: Just to check: is the `as` necessary here if we check the `instanceof` before it?

algochoi (commented):
Looks like a good improvement, just want to check a testcase for something like tuple(int)

tests/10.ABI.ts
michaeldiamant: @algochoi Good call - I added a test in https://github.com/algorand/js-algorand-sdk/pull/701/commits/cc4e928fda13e1b60d9aa10fe42d171984118bf0.  Can you resolve if it looks good?

Aside - Passing remarks:
* While adding the test, I discovered a bug in the test implementation.  I had inadvertently missed the return statement:  https://github.com/algorand/js-algorand-sdk/pull/701/commits/cc4e928fda13e1b60d9aa10fe42d171984118bf0#diff-002803be91ba5405e645aec5c397e490594a1883499139d9c69c070a594fd9dcR422.  I wish TypeScript treated if-statements as an expression.  Returning a value (like Scala) prevents this class of problem.
* I just realized the pre-commit hook will transform code.  I wrote _else if_ statements that got moved to if-statements.

tests/10.ABI.ts
michaeldiamant: @algochoi Thanks - You're correct.  The cast is not needed.  I simply didn't realize it, appreciate the pointer.  https://github.com/algorand/js-algorand-sdk/pull/701/commits/78f3e67243520c33581a35da365e72a71fbc2864 attempts to remedy the situation.

tests/10.ABI.ts
algochoi: Thanks, I think this looks good - resolving

algochoi (approved):

Issue #643, opened 2022-09-08
crypto module missing on react native app


davide99:
### Subject of the issue

I'm trying to use the sdk in a react native app. Since the react native doesn't include the node standard library the following error pops up:
The package at "node_modules/algosdk/dist/browser/algosdk.min.js" attempted to import the Node standard library module "crypto".

### Steps to reproduce

1. npx create-expo-app app
2. cd app
3. const algosdk = require('algosdk'); in a js file
4. npm start

Is there a way to fix this?
Thank you

jasonpaulos:
Unfortunately React Native is not a supported environment for this library.

Prior issues have been made that may or may not have helpful workarounds, e.g. https://github.com/algorand/js-algorand-sdk/issues/530

barnjamin:
Another dev ran into a similar issue and wrote up his experience here: https://medium.com/@shrinerp/algosdk-vue-js-webpack-and-node-js-d10014e87a55

Blackglade:
Getting this exact same issue with create-react-app as well.

```
Compiled with problems:X

ERROR in ./node_modules/algosdk/dist/browser/algosdk.min.js 7206:55-72

Module not found: Error: Can't resolve 'crypto' in '/Users/harshbaid/Desktop/algoranddao/node_modules/algosdk/dist/browser'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.
```

jasonpaulos:
> Getting this exact same issue with create-react-app as well.

@Blackglade you are seeing that because one of our dependencies, [tweetnacl-js](https://github.com/dchest/tweetnacl-js), uses `require('crypto')` to test if the Node.js `crypto` module is available. However, it is perfectly capable of running in a browser that does not have access to the `crypto` module, it just has to check which environment it's in, then fall back on browser functions if necessary.

This is add odds with your React webpack build, which thinks there must be a problem because it doesn't know about the `crypto` module. You must find some way of telling your React build that it's ok to not have that module. Consider modifying your webpack config to do this, possibly with the [`resolve.fallback`](https://webpack.js.org/configuration/resolve/#resolvefallback) option, e.g.:

```js
module.exports = {
  //...
  resolve: {
    fallback: {
      crypto: false, // do not include a polyfill for crypto
    },
  },
};
```

If you are aware of any better way this issue can be solved, please let us know.

[...]

PR #707, opened 2022-12-07
Packaging: Improve source map and browser usage for external bundlers

jasonpaulos:
While working on https://github.com/jasonpaulos/decipher-22-pyteal-talk, I became aware of a few deficiencies of using this SDK in React. This PR attempts to improve most, but not all, of the issues I encountered. They are:
* Missing sources: while we generate and ship source maps with our package, we don't include the actual sources, which make the source maps essentially useless since they have nothing to map to. To fix this, I've included the `src` directory in our release bundle.
  * As an additional improvement that wasn't strictly necessary, I moved the `index.ts` file into the `src` directory. This helps simplify the directory structure for our sources and compiled code. As far as I'm aware there's no downside to doing this.
* Failing to import the `crypto` module: as I explained [in this comment](https://github.com/algorand/js-algorand-sdk/issues/643#issuecomment-1241268534), our dependency [tweetnacl-js](https://github.com/dchest/tweetnacl-js) attempts to import the `crypto` module in all environments to see if it's available, but that module is not necessary in the browser. Due to the strict nature of bundlers, this commonly caused an error. However, by modifying the `browser` field in our `package.json`, we can declare that that module should be ignored by browser bundlers.
  * Note that [tweetnacl-js also does this](https://github.com/dchest/tweetnacl-js/blob/6a9594a35a27f9c723c5f1c107e376d1c65c23b3/package.json#L52-L55), which I thought would be sufficient, but for some reason is not.
* Building on Node v18: our version of webpack has a bug that prevents it from building on newer versions of node, so I updated it.
  * While I was updating packages I also updated chromedriver, but this isn't strictly necessary.

The other issue I found but is not addressed in this PR is with regard to `Buffer`, and I've described that more in #708.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,9 +44,7 @@ jobs:
     steps:
       - checkout
       - install_dependencies
-      - run: |
-          npm run build
-          npm run lint
+      - run: npm run lint
 
   prettier:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,9 @@ jobs:
     steps:
       - checkout
       - install_dependencies
-      - run: npm run lint
+      - run: |
+          npm run build
+          npm run lint
 
   prettier:
     docker:
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -48,8 +48,9 @@ module.exports = {
   ],
   ignorePatterns: [
     'dist/',
+    'docs/',
     'tests/cucumber/features/',
-    'tests/cucumber/build/',
+    'tests/cucumber/browser/build/',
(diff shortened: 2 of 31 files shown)


algochoi (commented):
Looks like good improvements:

> As an additional improvement that wasn't strictly necessary, I moved the index.ts file into the src directory. This helps simplify the directory structure for our sources and compiled code.

Just checking since the PR makes changes to local paths: this won't affect how external users use the module correct?

jasonpaulos:
That's correct, the path changes will not affect what's exposed to users. The files referenced from our `package.json` are unchanged.

algochoi (approved):
LGTM 🚢

PR #712, opened 2022-12-13
Remove langspec.json and logic over it

ahangsu:
Implements https://github.com/algorand/algorand-sdk-testing/issues/223

Part of https://github.com/algorand/js-algorand-sdk/issues/650


algochoi (approved):
Looks good, don't see any remaining references in source code

PR #715, opened 2022-12-15
Enhancement: Remove default parameters in AlgodClient constructor

barnjamin:
Rather than removing the port from the arguments as mentioned in https://github.com/algorand/js-algorand-sdk/pull/658#discussion_r978962110 , this pr just removes the default values. 

In terms of degrees of breakage, this seems like a lower severity break than removing port completely since currently functioning code will not break (as it would if port was taken out) but any new code would have a behavior  that more closely matches the devs expectation


michaeldiamant (approved):
@barnjamin The PR seems to me like a _reasonable_ compromise.

jasonpaulos (commented):
Seems like a good change, but I'd really like to see a test where something like `new AlgodClient("...", "https://example.com")` is confirmed to have the base URL `https://example.com` (and not `https://example.com:4180` like it would previously).

`tests/9.Client.ts` is probably a good place for this (though it seems like all the tests there work with the lower-level `URLTokenBaseHTTPClient` at the moment).

barnjamin:
@jasonpaulos good call, added a few tests to demonstrate how this should be functioning

jasonpaulos (approved):

PR #718, opened 2022-12-16
remove deprecated methods

ahangsu:


src/makeTxn.ts
```diff
@@ -64,53 +62,6 @@ export function makePaymentTxnWithSuggestedParams(
...
- * If the final calculated fee is lower than the protocol minimum fee, the fee will be increased to match the minimum.
- * @param amount - integer amount to send, in microAlgos
- * @param closeRemainderTo - optionally close out remaining account balance to this account, represented as string rep of Algorand address
- * @param firstRound - integer first protocol round on which this txn is valid
- * @param lastRound - integer last protocol round on which this txn is valid
- * @param note - uint8array of arbitrary data for sender to store
- * @param genesisHash - string specifies hash genesis block of network in use
- * @param genesisID - string specifies genesis ID of network in use
- * @param rekeyTo - rekeyTo address, optional
- * @Deprecated in version 2.0 this will change to use the "WithSuggestedParams" signature.
```
algochoi, line 81: @jasonpaulos Do we want to remove this function entirely? Or just change the signature so that it takes `WithSuggestedParams` as an argument?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Remove v1 algod API (`client.algod`) due to API end-of-life (2022-12-01). Instead, use v2 algod API (`client.v2.algod.algod`).
 - Remove `cost` field in `DryrunTxnResult` in favor of 2 fields: `budget-added` and `budget-consumed`. `cost` can be derived by `budget-consumed - budget-added`.
 - Remove logicsig templates (`logicTemplates`), `logic/langspec.json`, `logic.logic` depending on `langspec.json`.
+- Remove deprecated methods in `makeTxns.ts`, `logicsig.ts` and `types/transactions/encoded.ts`.
```
michaeldiamant, line 10: @ahangsu Minor:  Where possible, I propose the CHANGELOG guide the reader through the migration by explaining available recourses.

```suggestion
- Remove `makePaymentTxn` in favor of `makePaymentTxnWithSuggestedParams`.
- Remove `makeLogicSig` in favor of either using `LogicSigAccount` (preferred) or directly invoking `LogicSig` constructor.
- Remove `EncodedMultisigBlob` in favor of `EncodedSignedTransaction.
```

src/makeTxn.ts
michaeldiamant: Treat with a grain of salt because my JS knowledge is limited - From my read, there's a direct replacement via `makePaymentTxnWithSuggestedParams`.

So it _seems_ to me that the PR is accurate.

examples/logic_sig_example.js
```diff
@@ -1,6 +1,7 @@
 // Example: creating a LogicSig transaction signed by a program that never approves the transfer.
 
 const algosdk = require('../src');
+const { LogicSig } = require('../src/logicsig');
```
jasonpaulos, line 4: nit: for these examples we shouldn't import any library code that wouldn't otherwise be available to normal dependencies, i.e. what `require('algosdk')` gives you.

You could use `algosdk.LogicSig` here instead of importing the class directly. (Or better yet, `algosdk.LogicSigAccount`, if that's what we prefer people use.)

src/makeTxn.ts
jasonpaulos: The changes here look to be in line with the discussion we had today. I.e., removing the deprecated `makeX` functions in favor of the newer `makeXWithSuggestedParams`.

tests/7.AlgoSDK.js
```diff
@@ -1,5 +1,6 @@
 const assert = require('assert');
 const algosdk = require('../src/index');
+const { LogicSig } = require('../src/logicsig');
```
jasonpaulos, line 3: Similar to what I said about examples, we want our tests to import the code in a way that's as close as possible to what will happen in production. Meaning that we should prefer `algosdk.LogicSig` over importing internal files.

tests/8.LogicSig.ts
```diff
@@ -1,5 +1,6 @@
 import assert from 'assert';
 import algosdk from '../src/index';
+import { LogicSig } from '../src/logicsig';
```
jasonpaulos, line 3: Same here

tests/8.LogicSig.ts
```diff
@@ -433,7 +434,7 @@ describe('signLogicSigTransaction', () => {
 
   function testSign(
     lsigObject:
-      | ReturnType<typeof algosdk.makeLogicSig>
+      | ReturnType<(program: Uint8Array, args?: Uint8Array[]) => LogicSig>
```
jasonpaulos, line 437: `ReturnType<typeof algosdk.makeLogicSig>` was a way to get around the fact that `LogicSig` didn't used to be exported. Since it is now, I believe this can be replaced with `LogicSig` directly

ahangsu: 👍

tests/7.AlgoSDK.js
ahangsu: that makes sense

examples/logic_sig_example.js
ahangsu: 👍

src/makeTxn.ts
ahangsu: correct, I made the changes reflected in changelog for `make*Txn` in favor of `make*WithSuggestedParams`.

jasonpaulos (approved):

michaeldiamant (approved):

PR #720, opened 2022-12-16
v2: Regenerate models

algochoi:
This PR changes the generated constructors so that every constructor accepts an object instead of explicit arguments. Previously, this was only done for methods with greater than 4 arguments. 

Also deletes some stale types that have been since marked as `private` in the API spec.

Changes that stem from generator change: https://github.com/algorand/generator/pull/52/files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Remove v1 algod API (`client.algod`) due to API end-of-life (2022-12-01). Instead, use v2 algod API (`client.v2.algod.algod`).
 - Remove `cost` field in `DryrunTxnResult` in favor of 2 fields: `budget-added` and `budget-consumed`. `cost` can be derived by `budget-consumed - budget-added`.
 - Remove logicsig templates (`logicTemplates`), `logic/langspec.json`, `logic.logic` depending on `langspec.json`.
-- Regenerate algod models so every constructor requires an objects to be passed in. Previously, only constructors with more than 4 argument specified this.
+- Regenerate algod models so every constructor requires an object to be passed in. Previously, only constructors with more than 4 argument specified this.
 - Remove unused generated types: `CatchpointAbortResponse`, `CatchpointStartResponse`.
 
 # v1.24.1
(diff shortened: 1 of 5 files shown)


src/client/v2/algod/models/types.ts
```diff
@@ -440,11 +440,15 @@ export class AccountApplicationResponse extends BaseModel {
    * global data.
    * The raw account uses `AppParams` for this type.
    */
-  constructor(
-    round: number | bigint,
-    appLocalState?: ApplicationLocalState,
-    createdApp?: ApplicationParams
-  ) {
+  constructor({
```
michaeldiamant, line 443: @algochoi Unrelated to the line and minor:  Can you supply a CHANGELOG update to help the reader understand how to migrate?

algochoi: Thanks for the reminder, I keep forgetting this. I'll supply a commit here.

algochoi: [9b312db](https://github.com/algorand/js-algorand-sdk/pull/720/commits/9b312db44ccaff7a62d0ce8ef8180fe67da87b28)

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - Remove v1 algod API (`client.algod`) due to API end-of-life (2022-12-01). Instead, use v2 algod API (`client.v2.algod.algod`).
 - Remove `cost` field in `DryrunTxnResult` in favor of 2 fields: `budget-added` and `budget-consumed`. `cost` can be derived by `budget-consumed - budget-added`.
 - Remove logicsig templates (`logicTemplates`), `logic/langspec.json`, `logic.logic` depending on `langspec.json`.
+- Regenerate algod models so every constructor requires an objects to be passed in. Previously, only constructors with more than 4 argument specified this.
```
michaeldiamant, line 10:
```suggestion
- Regenerate algod models so every constructor requires an object to be passed in. Previously, only constructors with more than 4 argument specified this.
```

src/client/v2/algod/models/types.ts
michaeldiamant: @algochoi Resolving - Thanks for bringing in the CHANGELOG.

michaeldiamant (approved):
Matches my understanding for desired code generation.  Prefer a 2nd approval due to my limited JS knowledge.

jasonpaulos (approved):

PR #723, opened 2022-12-23
New Feature: able to generate wallets deterministically

palmerss:
Iv ran into a case where I need to be able to generate accounts in a deterministic way. I have seen other people keep a JSON file of Accounts before. That sounds like a pain, and it would be much cleaner to just have people come up with their own pre-determined seed generation and generate accounts using this function.

```
const deterministicAccounts = [];
for (let i = 0; i < 32; i++) {
  const bytesArr = new Uint8Array(new Array(32).fill(0));
  bytesArr.set(encodeUint64(i), 24);
  deterministicAccounts.push(algosdk.generateAccountFromSeed(bytesArr));
}
//deterministicAccounts now contains 32 accounts that are the same accounts every time I run my unit tests.
```

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -10,3 +10,12 @@ export default function generateAccount(): Account {
   const encodedPk = address.encodeAddress(keys.publicKey);
   return { addr: encodedPk, sk: keys.secretKey };
 }
+
+/**
+ * generateAccountFromSeed returns a new Algorand address and its corresponding secret key deterministically given a Uint8Array seed value
+ */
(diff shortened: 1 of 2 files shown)


algoanne:
Hi @palmerss thanks for the contribution, this would be valuable! I think this would be well-suited to the algokit-utils-ts repo, would you up for contributing it there? Could go nicely alongside the other account functions: https://github.com/algorandfoundation/algokit-utils-ts/blob/main/docs/capabilities/account.md#accounts

Issue #653, opened 2022-09-15
Allow manually adding foreign refs to transactions generated with `addMethodCall`


barnjamin:
In the case that a foreign ref (asset,acct,app) is not specified in the method signature but might be needed for some reason, the caller of the `AtomicTransactionComposer.addMethodCall` should still be able to provide it

ErikHasselwander:
I just want to add that this is a much needed feature, and the python SDK already has it. The workaround for this is very disruptive and requires you to break out of the ATC with `AtomicTransactionComposer.buildGroup()`.

barnjamin:
Re-upping this one 🙏

rocalex:
Same issue

ultrade-dev:
Guys can we please get this fixed?
It's disrupting our work.

PR #725, opened 2022-12-28
Enhancement: Add foreign array objects to ATC `addMethodCall`

algochoi:
This PR adds foreign array objects as arguments in `addMethodCall()` for the atomic transaction composer.

Adds a test that mirrors the Go SDK: https://github.com/algorand/go-algorand-sdk/pull/318. Currently written as a unit test, but could also benefit from a cucumber test since all the other SDKs implement this. 

Addresses #653

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -379,10 +379,12 @@ export class AtomicTransactionComposer {
     }
 
     const resolvedRefIndexes: number[] = [];
-    const foreignAccounts: string[] = appAccounts == null ? [] : appAccounts;
-    const foreignApps: number[] = appForeignApps == null ? [] : appForeignApps;
+    const foreignAccounts: string[] =
+      appAccounts == null ? [] : appAccounts.slice();
+    const foreignApps: number[] =
+      appForeignApps == null ? [] : appForeignApps.slice();
     const foreignAssets: number[] =
-      appForeignAssets == null ? [] : appForeignAssets;
+      appForeignAssets == null ? [] : appForeignAssets.slice();
     for (let i = 0; i < refArgTypes.length; i++) {
       const refType = refArgTypes[i];
       const refValue = refArgValues[i];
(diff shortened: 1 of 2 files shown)


src/composer.ts
```diff
@@ -370,9 +379,10 @@ export class AtomicTransactionComposer {
     }
 
     const resolvedRefIndexes: number[] = [];
-    const foreignAccounts: string[] = [];
-    const foreignApps: number[] = [];
-    const foreignAssets: number[] = [];
+    const foreignAccounts: string[] = appAccounts == null ? [] : appAccounts;
+    const foreignApps: number[] = appForeignApps == null ? [] : appForeignApps;
```
barnjamin, line 383: should this be `=== undefined`? I'm actually not sure

algochoi: I think that's fine too, `== null` is a looser check used later in this file so I followed the previous conventions here: https://github.com/algorand/js-algorand-sdk/pull/725/files#diff-534b9601ad1de4bb83d07dc0cb5af85424ccf602bcd8f4c0a2d3b79def06e8aeR458

barnjamin (approved):

src/composer.ts
```diff
@@ -370,9 +379,10 @@ export class AtomicTransactionComposer {
...
+    const foreignAssets: number[] =
+      appForeignAssets == null ? [] : appForeignAssets;
```
jasonpaulos, line 385: Can you make a copy of these input arrays (e.g. `appAccounts.slice()`)?

That way we will avoid modifying inputs

jasonpaulos (commented):
Thanks for working on this, it's a much needed improvement. I found one small issue but this looks good to merge after that's addressed.

I agree that we need a new cucumber test to ensure all SDKs have appropriate coverage here, but I don't think that's a blocker for this PR.

src/composer.ts
algochoi: Good point, made a copy in commit: [c918bc0](https://github.com/algorand/js-algorand-sdk/pull/725/commits/c918bc0a220ed7e1fdd2c3730528f8ec240cd9a1)

jasonpaulos (approved):

Issue #708, opened 2022-12-07
Explicitly import `Buffer`


jasonpaulos:
Right now our code does not explicitly import any package to use `Buffer`; we just reference it as a global variable.

This is acceptable in Node, but browsers don't have a `Buffer` implementation. To solve this, [we use a webpack plugin](https://github.com/algorand/js-algorand-sdk/blob/f7ea10a2cfbe0f7f167b46e1612f0fbee4dd7d6f/webpack.config.js#L20-L24) to provide `Buffer` as a global variable.

This works well for us, but when users attempt to use our package with a browser framework like React that does its own bundling, their build process doesn't know about this and will fail out of the box. Users are forced to change their own bundler configuration to match the change we made for `Buffer`. And since React apps are often started with create-react-app, which requires you to eject from the initial setup to modify bundler configs, this process has a lot of friction.

To improve things, we could explicitly import `Buffer` everywhere in the codebase where we reference it so that modifying our webpack config is no longer necessary. Some care will need to be taken to ensure that Node still uses the native `Buffer` implementation.

AustP:
Our react apps currently use algosdk v1.20.0 and we were able to expose Buffer on the window object right as the app started up as a workaround.

```js
window.Buffer = require('buffer/').Buffer;
```

After updating algosdk to v1.24.1, this workaround no longer works. I'm not sure what changed between the versions, but looking forward to being unblocked when the imports are fixed for Buffer.

barnjamin:
@AustP I made a little CRA app to test and had to add the following to the `config-overrides.js` after setting up react-app-rewired

```js
const webpack = require('webpack');
module.exports = function override(config, env) {
    config.plugins.push(new webpack.ProvidePlugin({Buffer: ['buffer', 'Buffer']}))
    return config
}
```

PR #730, opened 2023-01-09
Update uint64.ts

headline-design:
Fix for error 'buf.writeBigUInt64BE is not a function,' which suddenly emerged across numerous bundlers and prevents browsers from using any algosdk methods that utilize encodeUint64 (such as algosdk.getApplicationAddress).

### DIFF
--- a/src/encoding/uint64.ts
+++ b/src/encoding/uint64.ts
@@ -5,18 +5,38 @@
  * @returns An 8-byte typed array containing the big-endian encoding of the input
  *   integer.
  */
-export function encodeUint64(num: number | bigint) {
-  const isInteger = typeof num === 'bigint' || Number.isInteger(num);
+export function encodeUint64(dec: number) {
+  let binString = convertToBinary(dec)
 
-  if (!isInteger || num < 0 || num > BigInt('0xffffffffffffffff')) {
-    throw new Error('Input is not a 64-bit unsigned integer');
+  while (true) {
+    if (binString.length < 64) {
+      binString = "0" + binString
+    }
+    else {
+      break
+    }
   }
 
-  const buf = Buffer.allocUnsafe(8);
+  let array : number[] = [];
 
-  buf.writeBigUInt64BE(BigInt(num));
+  for (let i = 0; i < 64; i += 8) {
+    let piece = binString.slice(i, i + 8)
+    piece += ""
+    let pieceNumber = parseInt(piece, 2)
+    array.push(pieceNumber)
+  }
 
-  return new Uint8Array(buf);
+  return Uint8Array.from(array)
+
+}
+
+function convertToBinary (num: number) {
+  let binary = (num % 2).toString();
+  for (; num > 1; ) {
+      num = parseInt(String(num / 2));
+      binary =  (num % 2) + (binary);
+  }
+  return binary
 }
 
 /**


jasonpaulos:
Hi @headline-design, before we introduce a fix I'm interested in learning more about why this problem occurred in the first place. Can you tell me more about the environment in which you're seeing the problem, specifically:
1. Which version of the SDK are you using?
2. Are you using a browser framework like React or Vite? If so, which version, and have you modified any bundler properties to make this SDK work?
3. Which browser(s) does this error occur in?
4. Since you said this function used to work fine, are any changes made since then that might account for the difference?
5. Could you run `npm ls buffer` from the root of your project and let me know the output?

headline-design:
> Which version of the SDK are you using?

Fails on all tested versions including previously working version around 1.17

> Are you using a browser framework like React or Vite? 

Yes. Fails on both React and Parcel (all versions back to 17).  Of note: no error occurs when using the algosdk in nodeJS, solely when attempting to bundle and run code in browser.

> If so, which version, and have you modified any bundler properties to make this SDK work?

No. You can verify that the error occurs by creating a blank React App with `npx create-react-app TestApp` and running the following code in App.js:

`
import algosdk from 'algosdk'

console.log(algosdk.encodeUint64(123456))
console.log(algosdk.getApplicationAddress(123456))
`

> Which browser(s) does this error occur in?

Firefox and Chrome

> Since you said this function used to work fine, are any changes made since then that might account for the difference?

No changes made on our end. 

>     5. Could you run `npm ls buffer` from the root of your project and let me know the output?

Done. Both functions still fail in a bare-bones React app. Result:

`-- algosdk@2.0.0
  `-- buffer@6.0.3

browser error: Buffer is not defined

headline-design:
Reference to this issue in topic on Forum: https://forum.algorand.org/t/encodeuint64-asset-id-doesnt-work/8572

jasonpaulos:
Unfortunately I'm not able to reproduce this error like you described. I used `create-react-app` to create a new app, installed `algosdk`, and tried to run the following code:

```javascript
console.log(algosdk.encodeUint64(123456))
console.log(algosdk.getApplicationAddress(123456))
```

The app fails with this error:

```
Uncaught ReferenceError: Buffer is not defined
    at ./node_modules/algosdk/dist/esm/encoding/address.js (address.ts:30:1)
    at options.factory (react refresh:6:1)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:62:1)
    at ./node_modules/algosdk/dist/esm/main.js (logicsig.ts:512:1)
    at options.factory (react refresh:6:1)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:62:1)
    at ./node_modules/algosdk/dist/esm/index.js (group.ts:97:1)
    at options.factory (react refresh:6:1)
```

Until we fix #708, this is unfortunately expected.

@headline-design do these steps to reproduce the issue match what you described? Would it be possible for you to create a github repo that reproduces this error?

I'd like to figure out why this is happening but unfortunately I can't do that without reproducing it locally.

headline-design:
Certainly: We have two version with two different errors resulting from the same root issue for you to consider: Here is the github repo where the error can be replicated: https://github.com/headline-design/algosdkBufferError. Here is a deployed version where the error can be seen in the console.log - https://algosdk-buffer-error.vercel.app/.

Note: The example is a standard Create-react-app. However, the error changes depending on the version of 'react-scripts' that is installed. for version > 4 the error reads: "buff.write bigUint64...".  'React scripts' > 5 the error reads "Buffer is not defiend."

Note: The current error repo is for create-react-app with React Scripts 4. The current deployed vercel is with React Scripts 5. Both version are built with js-algorand-sdk 2.0.

headline-design:
So just to clarify that explicitly adding Buffer will not fix this issue. Buffer no longer supports the old (standard) version of BigUint64.

jasonpaulos:
Thank you! I was able to reproduce the error from your github repo. I'll investigate further as to why this is happening

jasonpaulos:
So here's what I believe is happening: `react-script` v4 uses `webpack` v4, which includes `node-libs-browser`, which makes a lot of node features available in the browser, including `Buffer`. The problem is that it uses an older version of the `buffer` package (4.9.2) which doesn't have the `writeBigUInt64BE` function.

You can see this by running this command:

```
$ npm ls buffer
testbuffer@0.1.0 /Users/jason/Workspace/algosdkBufferError
├─┬ algosdk@2.0.0
│ └── buffer@6.0.3
└─┬ react-scripts@4.0.3
  └─┬ webpack@4.44.2
    └─┬ node-libs-browser@2.2.1
      └── buffer@4.9.2
```

With no additional modifications to your bundler settings, `buffer@4.9.2` is the one that gets added to `window.Buffer` and consumed by `algosdk`.

If you've ever wondered why `algosdk` can't even find Buffer when using `react-scripts` v5, it's because they now use `webpack` v5, which no longer includes `node-libs-browser`.

A long term solution is for us to work on #708, which make our `buffer` dependency behave like a normal dependency instead of relying on whatever `window.Buffer` happens to be.

However, if you're looking for a short term workaround (and you use npm v8.3.0 or higher), you could use the [overrides](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) property in your `package.json` to override the version of `buffer` that gets installed by `node-libs-browser`.

For example, by adding this to the `package.json` of https://github.com/headline-design/algosdkBufferError:

```javscript
  "overrides": {
    "buffer": "6.0.3"
  }
```

I now get:

```
$ npm ls buffer
testbuffer@0.1.0 /Users/jason/Workspace/algosdkBufferError
├─┬ algosdk@2.0.0
│ └── buffer@6.0.3 overridden
└─┬ react-scripts@4.0.3
  └─┬ webpack@4.44.2
    └─┬ node-libs-browser@2.2.1
      └── buffer@6.0.3 deduped
```

And the app works!

jasonpaulos:
I've just opened #733 which should solve the problem.

I've verified this locally, but @headline-design could you check as well? You should be able to trial the change by doing:

```
npm uninstall algosdk
npm install --save algorand/js-algorand-sdk#pull/733/head
```

jasonpaulos:
#733 was just merged, so I'm closing this in favor of that

PR #735, opened 2023-01-16
API: verifyTealSign utility function

M-Picco:
Introduce a new function `verifyTealSign` which can be used to verify the output of its pair method, `tealSign`.

Some applications might benefit from having this utility function, either for testing purposes or not, and it would save developers the work of importing the `nacl` library and doing the verification manually by themselves.

### DIFF
--- a/src/logicsig.ts
+++ b/src/logicsig.ts
@@ -496,6 +496,29 @@ export function tealSign(
 }
 
 /**
+ * verifyTealSign verifies a signature as would the ed25519verify opcode
+ * @param sk - uint8array with public key to verify against
+ * @param data - buffer with original signed data
+ * @param programHash - string representation of teal program hash (= contract address for LogicSigs)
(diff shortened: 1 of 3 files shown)


src/logicsig.ts
```diff
@@ -495,6 +495,29 @@ export function tealSign(
   return nacl.sign(toBeSigned, sk);
 }
 
+/**
+ * verifyTealSign verifies a signature as would the ed25519verify opcode
+ * @param sk - uint8array with public key to verify against
```
barnjamin, line 500: I think this should be `pk` and at the bottom?

M-Picco: You are right. I'll fix it

bbroder-algo:
I don't have any problem with this nacl helper function, and LGTM, but I am not particularly qualified for JS.

jasonpaulos (approved):
This is an appreciated change. I believe we intended to add this at one point, but we never got around to it. Implementation and tests look good, as long as it passed CI.

Issue #638, opened 2022-09-01
Typescript error on deployment to AWS


berkaey:
I'm getting this error while deploying to AWS ElasticBeanstalk environment. BTW, In local, it works perfectly.
```
node_modules/algosdk/dist/types/src/types/utils.d.ts(19,84): error TS2344: Type 'K' does not satisfy the constraint 'Partial<T>'.
```

Local Environment
```
TS: 4.6.3
Node: 16.4
SDK: 1.19.1
```

EB Environment
```
TS: 4.6.3
Node: 16 running on 64bit Amazon Linux 2/5.5.5 (arm64)
SDK: 1.19.1
```

mploux:
Exact same problem here, but it is during build time. Not specific to an AWS deployment

barnjamin:
@berkaey and @mploux do you have any more info on this? a longer stack trace even?

berkaey:
I'm not sure it is directly related to algo-sdk because after I forced ts version to 4.7.4 and removing node_modules from tsconfig paths the error disappeared

hernandp:
Same problem, output:

```
node_modules/algosdk/dist/types/src/types/utils.d.ts:19:84 - error TS2344: Type 'K' does not satisfy the constraint 'Partial<T>'.

19 export declare type DistributiveOverwrite<T, K> = T extends unknown ? Overwrite<T, K> : never;
                                                                                      ~

  node_modules/algosdk/dist/types/src/types/utils.d.ts:19:46
    19 export declare type DistributiveOverwrite<T, K> = T extends unknown ? Overwrite<T, K> : never;
                                                    ~
    This type parameter might need an `extends Partial<T>` constraint.

```

hernandp:
UPDATE:  Upgrade to Typescript 4.7.4, error gone.

Vidhyanandcs:
i am using typescript 4.8.2
Got the same error...

`node_modules/algosdk/dist/types/src/types/utils.d.ts:19:84 - error TS2344: Type 'K' does not satisfy the constraint 'Partial<T>'.

19 export declare type DistributiveOverwrite<T, K> = T extends unknown ? Overwrite<T, K> : never;
                                                                                      ~

  node_modules/algosdk/dist/types/src/types/utils.d.ts:19:46
    19 export declare type DistributiveOverwrite<T, K> = T extends unknown ? Overwrite<T, K> : never;
    This type parameter might need an `extends Partial<T>` constraint.`

[...]

PR #741, opened 2023-02-12
bugfix: Satisfy typescript constraints

aorumbayev:
## Description

I was using js algosdk as part of a ts npm package that relies on https://www.npmjs.com/package/dts-bundle-generator for bundle generation. The main dependency was beaker-ts (which relies on algosdk) client however dts bundler revealed tiny ts related errors that had to be addressed to properly transpile the code. Refer to screens below describing the errors that were addressed.

## Proposed changes

1. Error in DistributiveOverwrite in utils.ts file
screen1

2. Error in ConstructTransaction in builder.ts file
screen2

## Further notes 

I will rely on the forked version of the repo in the meantime

### DIFF
--- a/src/types/transactions/builder.ts
+++ b/src/types/transactions/builder.ts
@@ -45,10 +45,10 @@ export type TransactionBase =
  * - A: Additional parameters on top of the base transaction parameters
  * - O: A set of overwrites for transaction parameters
  */
-export type ConstructTransaction<A = {}, O = {}> = DistributiveOverwrite<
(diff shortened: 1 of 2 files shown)


barnjamin:
related issue https://github.com/algorand/js-algorand-sdk/issues/638

algochoi (approved):
Thank you for the change - I can confirm that I get similar suggestions on my editor and this looks like the most straightforward fix to me.

jasonpaulos (approved):

PR #746, opened 2023-03-08
Simulate unsigned transactions

algochoi:
This PR allows unsigned transactions to be encoded and sent to the `simulate` endpoint. 

Also implements tests defined in https://github.com/algorand/algorand-sdk-testing/pull/270

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -1,32 +1,32 @@
 import { Buffer } from 'buffer';
 import {
-  ABIType,
-  ABITupleType,
-  ABIUintType,
   ABIAddressType,
-  ABIValue,
+  abiCheckTransactionType,
   ABIMethod,
   ABIReferenceType,
-  abiTypeIsTransaction,
-  abiCheckTransactionType,
+  ABITupleType,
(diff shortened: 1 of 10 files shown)


src/composer.ts
```diff
@@ -1,32 +1,32 @@
 import { Buffer } from 'buffer';
 import {
-  ABIType,
```
algochoi, line 3: Used VSCode's `Organize Import`

tests/cucumber/steps/steps.js
```diff
@@ -459,18 +459,13 @@ module.exports = function getSteps(options) {
     async function (amt, note) {
       [this.pk] = this.accounts;
       const result = await this.v2Client.getTransactionParams().do();
-      this.lastRound = result.lastRound;
-      this.txn = {
+      this.txn = algosdk.makePaymentTxnWithSuggestedParamsFromObject({
```
algochoi, line 462: Uses the canonical factory method to build up the `Transaction` object instead of relying on duck typing to infer the object type.

PR #747, opened 2023-03-16
Docs: add examples to be pulled in for docs

joe-p:
Removed all the old examples and updated with new examples with tags for the new docs refactor.


bbroder-algo (approved):
probably works on my machine

PR #754, opened 2023-03-23
Docs: Examples

barnjamin:
Tweaks, add smoke test to CI


bbroder-algo (approved):
don't like the name smoke-test

Issue #756, opened 2023-03-25
KeyReg transaction typing forces value of NonParticipating flag


barnjamin:
### Subject of the issue

A Keyreg transaction to take an account offline can have the non participating flag set to true or false

https://github.com/algorand/go-algorand/blob/master/ledger/apply/keyreg.go#L58-L66

Our SDK forces it to be true or false depending which overload type is used

https://github.com/algorand/js-algorand-sdk/blob/develop/src/makeTxn.ts#L205

We should not do that

PR #757, opened 2023-03-25
Fix: allow nonParticipation flag to be a boolean

barnjamin:
addresses #756

### DIFF
--- a/src/makeTxn.ts
+++ b/src/makeTxn.ts
@@ -200,10 +200,8 @@ export function makeKeyRegistrationTxnWithSuggestedParamsFromObject(
         'reKeyTo',
         'rekeyTo'
       >,
-      'from' | 'note' | 'suggestedParams' | 'rekeyTo'
-    > & {
-      nonParticipation: boolean;
-    }
+      'from' | 'note' | 'suggestedParams' | 'rekeyTo' | 'nonParticipation'
+    >
   >
 ): txnBuilder.Transaction;
 export function makeKeyRegistrationTxnWithSuggestedParamsFromObject(o: any) {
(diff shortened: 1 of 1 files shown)


src/makeTxn.ts
```diff
@@ -202,7 +202,7 @@ export function makeKeyRegistrationTxnWithSuggestedParamsFromObject(
       >,
       'from' | 'note' | 'suggestedParams' | 'rekeyTo'
     > & {
-      nonParticipation: true;
+      nonParticipation: boolean;
     }
```
jasonpaulos, line 206: You can simplify it to this:

```js
  o: Expand<
    Pick<
      RenameProperty<
        MustHaveSuggestedParams<KeyRegistrationTxn>,
        'reKeyTo',
        'rekeyTo'
      >,
      'from' | 'note' | 'suggestedParams' | 'rekeyTo' | 'nonParticipation'
    >
  >
```

Note this indicates a _slightly_ different meaning, since if `nonParticipation` is not included, it defaults to `false`, whereas in your original change it would be a required parameter

jasonpaulos (commented):
The problem it solves is that the SDK did not allow you to create offline keyreg transactions, let's be clear about that.

I've verified locally that this works, with a small tweak.

src/makeTxn.ts
barnjamin: makes sense,  thanks!

jasonpaulos (approved):
Looks good.

Note we'd have discovered this problem earlier if `tests/5.Transaction.js` was a typescript file. It would be a good idea to convert this and the other JS test files to TS.

algochoi (approved):

PR #760, opened 2023-04-05
BugFix: min fee sp fix

### LINKED ISSUE #750: Missing `min-fee` on Suggested params
### Subject of the issue

Other SDKs return the `min-fee` returned in the response from the algod transaction params endpoint:
https://github.com/algorand/py-algorand-sdk/blob/develop/algosdk/v2client/algod.py#L443

The JS SDK is missing this field:
https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/v2/algod/suggestedParams.ts#L20

barnjamin:
Closes #750


src/types/transactions/builder.ts
```diff
@@ -50,12 +50,14 @@ export type ConstructTransaction<
   O extends Partial<TransactionBase & A> = {}
 > = DistributiveOverwrite<TransactionBase & A, O>;
 
+export type SuggestedParamsWithoutMinFee = Omit<SuggestedParams, 'minFee'>;
```
jasonpaulos, line 53: I think it would be more backwards-compatible to do the reverse: introduce a new `SuggestedParamsWithMinFee` that `SuggestedParamsRequest` would produce. That way existing code referencing `SuggestedParams` will remain the same.

Note that this would only work if a `SuggestedParamsWithMinFee` can be viewed as a `SuggestedParams`,  which I think should work because the former has all the fields required of the latter, but we'd need to double check to be sure.

jasonpaulos (approved):
Looks good, thanks for taking my suggestion.

Just one inconsequential nit

algochoi (approved):
Looks good, like the backward compatible change

PR #768, opened 2023-04-24
Simulation: Lift log limits option in SimulateRequest

ahangsu:


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="simulate-request-log-option"
```
ahangsu, line 3:
```suggestion
SDK_TESTING_BRANCH="master"
```

src/composer.ts
```diff
@@ -615,7 +621,8 @@ export class AtomicTransactionComposer {
    *   in this group (ABIResult[]) and the SimulateResponse object.
    */
   async simulate(
-    client: Algodv2
+    client: Algodv2,
+    request?: SimulateRequest
```
jasonpaulos, line 625: Can you add this param to the docs above with similar wording as the python SDK?

src/composer.ts
```diff
@@ -627,8 +634,24 @@ export class AtomicTransactionComposer {
...
+
+    const currentRequest: SimulateRequest = new SimulateRequest({
+      txnGroups: [
+        new SimulateRequestTransactionGroup({
+          txns: txnObjects,
+        }),
+      ],
+      allowMoreLogging: request == null ? false : request.allowMoreLogging,
+      allowEmptySignatures:
+        request == null ? false : request.allowEmptySignatures,
```
jasonpaulos, line 649: Since we know we want to add more properties to `SimulateRequest` in the future and it is a generated class here, it would be great to avoid having to manually reference all the properties here

tests/cucumber/steps/steps.js
```diff
@@ -4714,6 +4714,45 @@ module.exports = function getSteps(options) {
     }
   );
 
+  When('I make a new simulate request.', async function () {
+    this.simulateRequest = new algosdk.modelsv2.SimulateRequest(
+      [],
```
jasonpaulos, line 4719: I don't think this is right. `SimulateRequest` requires one argument that's an object. Maybe it happens to work because of JS soft typing, but I'd expect this to be `SimulateRequest({ txnGroups: [] })`

ahangsu: ok, type script actually screws me up 😢

jasonpaulos: The problem is this file isn't typescript...

src/composer.ts
ahangsu: see [pr comment response](https://github.com/algorand/js-algorand-sdk/pull/768/commits/fe95c4cbd087bffbfe7703092a28e0936a0e2186)

tests/cucumber/steps/steps.js
ahangsu: see [pr comment response](https://github.com/algorand/js-algorand-sdk/pull/768/commits/fe95c4cbd087bffbfe7703092a28e0936a0e2186)

src/composer.ts
jasonpaulos: That looks good!

jasonpaulos (approved):
Once you revert `.test-env`, looks good to me!

Issue #654, opened 2022-09-17
Breaking changes in v3.0.0


michaeldiamant:
Itemizes breaking changes slated for the next major version bump.

PR #769, opened 2023-04-24
algod: Add blockOffsetTimestamp, syncRound APIs to algod client

Eric-Warehime:
Updates the algod client to add support for the following new APIs:
* /v2/ledger/sync/{round} POST
* /v2/ledger/sync/ GET
* /v2/devmode/blocks/offset/{offset} POST
* /v2/devmode/blocks/offset GET
* /v2/ledger/sync GET
* /v2/ledger/sync/{round} POST
* /v2/ledger/sync DELETE
* /ready

Also incorporates generated types into the model from the algod spec.


src/client/v2/algod/getBlockOffsetTimestamp.ts
```diff
@@ -0,0 +1,8 @@
+import JSONRequest from '../jsonrequest';
+
+export default class GetBlockOffsetTimestamp extends JSONRequest {
```
jasonpaulos, line 3: I know these are relatively simple endpoints, but I highly recommend unpacking the responses into the generated model types for new endpoints. That way the user will see the response as a typed object (`GetBlockTimeStampOffsetResponse` in this case), instead of a generic `Record<string, any>` value.

To do this, you can inherit from `JSONRequest<ModelType, Record<string, any>>`, and implement the `prepare(body: Record<string, any>): ModelType` method. See how `GetApplicationBoxes` does this here: https://github.com/algorand/js-algorand-sdk/blob/a89046acaf9502f496fca9d781bf33047c969c01/src/client/v2/algod/getApplicationBoxes.ts#L57-L60

Really _all_ of our endpoints should do this, but for backwards compatibility we have not changed endpoints that existed before this behavior was available.

algochoi (commented):
Is there going to be support for the unset sync endpoint in the SDKs?

Let's also add the `/ready` endpoint as well

src/client/v2/algod/getBlockOffsetTimestamp.ts
Eric-Warehime: Sounds good I'll make the change.

I'm also going to add an issue for that in #654 so that we can track these changes for the next time we want to make breaking changes.

Eric-Warehime: Issue for posterity: https://github.com/algorand/js-algorand-sdk/issues/771

tests/cucumber/integration.tags
```diff
@@ -13,4 +13,3 @@
 @rekey_v1
 @send
 @send.keyregtxn
-@simulate
```
jasonpaulos, line 16: @ahangsu just making you aware that you'll need to add back this tag in the simulate PR you're working on

ahangsu: sure sure, it is on my stack.

jasonpaulos (approved):
Pending tests, LGTM

algochoi (approved):
👍

PR #773, opened 2023-05-02
node: Drop support for node v14, npm audit fix

Eric-Warehime:
Updates node version minimum and in circle CI to avoid relying on a deprecated version.

Also ran `npm audit fix` to update dependencies in order to get fixes for vulnerabilities in our dependency packages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,7 +127,7 @@ commands:
             << parameters.sudo >> apt -y update
             << parameters.sudo >> apt -y install curl make git build-essential jq unzip
       - node/install:
-          node-version: '14'
+          node-version: '16'
       - run:
           name: npm ci
           command: |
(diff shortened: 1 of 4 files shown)


winder (approved):
Seems fine

algochoi (approved):
This looks necessary to fix the CI tests - thanks

jasonpaulos (approved):
Changes look good, but can you rename the title to more clearly indicate we're dropping support for node 14? This will affect users, so it should be clear in our changelog

Eric-Warehime:
> I also don't agree about categorizing this as a bug fix

The bugfix label is referring to the security vulnerabilities fixed by `npm audit fix`. 

Which label would you prefer out of:
`"New Feature, Enhancement, Bug-Fix, Not-Yet-Enabled, Skip-Release-Notes"`

jasonpaulos:
I think Enhancements is a better descriptor, the reason being putting a minimum node version change under the "Bugfixes" header of our changelog does not seem right.

Thanks for the title change

Issue #635, opened 2022-08-26
Skip preflight queries for API endpoints with no tokens


urtho:
## Problem

Any usage of custom headers in HTTP requests causes browsers to send extra preflight query.
This makes no sense in case of empty token and can be avoided by not sending the header at all.

## Solution

Do not attach 'X-Indexer-API-Token' | 'X-KMD-API-Token' | 'X-Algo-API-Token' if case of empty header value.

urtho:
Thought this would make it into v2.0.0 :'(

Eric-Warehime:
@urtho 

It seems like there's maybe two different types of issues here:

1. Algod and Kmd both require a token for almost all of their APIs (except for things like `/versions`, /health`, `/ready`, etc). As far as I know you can't run these APIs without an authenticator which checks a provided token.

 We could mark certain endpoints as part of the set of [common APIs](https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/server/common/routes.go#L24) and avoid sending the headers for these endpoints.


2. It's possible to run an indexer without auth, by not providing the [tokenString](https://github.com/algorand/indexer/blob/develop/cmd/algorand-indexer/daemon.go#L304). In this case you probably wouldn't want to send the header ever.

Are you interested in both of these cases or just one in particular?

urtho:
My use case is only for front-end apps where token (client side) is assigned an empty value ('', null, undefined) in the config. 
My ask is to not send a token header (indexer & algod) at all in this and only this case.

So I am interested in 1&2 - but they do not require a grouping APIs as the token is empty (client side) either way.
This should not break things as not sending a header vs sending one with empty value is of no consequence to algod / indexer.

Providers hide algod/indexer behind gateways that can choose to do different auth mech - like JWT or URL based tokens.
Anything but a custom header that triggers preflight for each request doubling the response latency.

Eric-Warehime:
PR #782 is merged. It should go out in the next SDK release (sometime in the next couple of weeks I'd guess?). Feel free to re-open this if you pick up this change and it doesn't work for you.

PR #782, opened 2023-05-12
Client: Don't send auth header when token is empty

Eric-Warehime:
Implements #635 

This change removes the custom auth headers from requests for all of kmd, algod, and indexer clients. 

After removing this we should be sending only standard headers in requests which should avoid the preflight queries mentioned in 635.

I've added cucumber tests in https://github.com/algorand/algorand-sdk-testing/pull/282 and one test in `tests/9.Client.ts`

### DIFF
--- a/src/client/v2/serviceClient.ts
+++ b/src/client/v2/serviceClient.ts
@@ -19,6 +19,9 @@ function convertTokenStringToTokenHeader(
   headerIdentifier: TokenHeaderIdentifier
 ): TokenHeader {
   const tokenHeader = {};
+  if (token === '') {
+    return tokenHeader
+  }
   tokenHeader[headerIdentifier] = token;
   return tokenHeader as TokenHeader;
 }
--- a/src/client/v2/serviceClient.ts
+++ b/src/client/v2/serviceClient.ts
@@ -20,7 +20,7 @@ function convertTokenStringToTokenHeader(
 ): TokenHeader {
   const tokenHeader = {};
   if (token === '') {
-    return tokenHeader
+    return tokenHeader;
   }
   tokenHeader[headerIdentifier] = token;
   return tokenHeader as TokenHeader;
--- a/tests/9.Client.ts
+++ b/tests/9.Client.ts
@@ -161,6 +161,12 @@ describe('client', () => {
       assert.strictEqual(client.c['bc']['baseURL']['port'], '456');
     });
 
+    it('should not provide auth request headers when the token is empty', () => {
+      const client = new AlgodClient('', `${baseServer}:${123}`, 456);
+      assert.deepStrictEqual({...client.c['bc']['tokenHeaders'], ...client.c['bc']['defaultHeaders']}, {})
+
+    })
+
     /* eslint-disable dot-notation */
   });
 });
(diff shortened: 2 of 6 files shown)


src/client/v2/serviceClient.ts
```diff
@@ -19,6 +19,9 @@ function convertTokenStringToTokenHeader(
   headerIdentifier: TokenHeaderIdentifier
 ): TokenHeader {
   const tokenHeader = {};
+  if (token === '') {
```
tzaffi, line 22: Is it clear that in the issue being encountered the empty token provided is of type `string`? I see other possibilities such as `AlgodTokenHeader` so if these were being provided empty, then the fix wouldn't apply.

tzaffi: NVM. I see now that we can only get here in the case of `string`: https://github.com/algorand/js-algorand-sdk/blob/5f8097fe2bfcb6b61158d8702c28ad43ca79742c/src/client/v2/serviceClient.ts#L55

tests/9.Client.ts
```diff
@@ -161,6 +161,17 @@ describe('client', () => {
       assert.strictEqual(client.c['bc']['baseURL']['port'], '456');
     });
 
+    it('should not provide auth request headers when the token is empty', () => {
```
tzaffi, line 164: nit: test kmd and indexer clients also?

tzaffi (commented):
Looks good.
I recommend modding [the .test-env](https://github.com/algorand/js-algorand-sdk/blob/933545befc8f88e9aa6baf97040704f56d9b522d/.test-env#L3) to temporarily re-point

```env
STD_TESTING_URL="https://github.com/Eric-Warehime/algorand-sdk-testing"
STD_TESTING_BRNACH="client-no-header-tests"
```
so we can see the tests run in CI.

tests/cucumber/steps/steps.js
```diff
@@ -1604,6 +1604,28 @@ module.exports = function getSteps(options) {
     );
   });
 
+  Given('expected headers', (headersTable) => {
+    this.expectedHeaders = headersTable.hashes();
```
jasonpaulos, line 1608: I'm guessing that `headersTable` is an instance of a class that has the `hashes` method. Since JSON serialization + deserialization cannot reconstruct class instances, only basic objects, those methods are lost during browser testing when arguments are passed through the webdriver protocol.

That's likely why the tests are failing in chrome + firefox.

Eric-Warehime: Yep, I'm taking a look now...if you have any suggestions on how to handle this let me know. It's pretty unfortunate that the browser tests seem to prevent us from using some builtin features of cucumber.

Eric-Warehime: Added the test you suggested.

Eric-Warehime:
> I recommend modding [the .test-env](https://github.com/algorand/js-algorand-sdk/blob/933545befc8f88e9aa6baf97040704f56d9b522d/.test-env#L3)

Here is the link to the passing CI w/ test-env pointed at the altered cucumber tests. https://app.circleci.com/pipelines/github/algorand/js-algorand-sdk/1040/workflows/8182c1e1-ef40-4841-bd35-e2163b496ff4

jasonpaulos (approved):

PR #785, opened 2023-05-18
algod: Add state delta APIs

Eric-Warehime:
Adds new APIs to the algod client:
* `/v2/deltas/{round}`
* `/v2/deltas/{round}/txn/group`
* `/v2/deltas/txn/group/{id}`

Implements the new cucumber tests for the above as well.

### DIFF
--- a/src/client/v2/algod/getLedgerStateDeltaForTransactionGroup.ts
+++ b/src/client/v2/algod/getLedgerStateDeltaForTransactionGroup.ts
@@ -1,0 +1,15 @@
+import JSONRequest from '../jsonrequest';
+import HTTPClient from '../../client';
+import IntDecoding from '../../../types/intDecoding';
+
+export default class GetLedgerStateDeltaForTransactionGroup extends JSONRequest {
+  constructor(c: HTTPClient, intDecoding: IntDecoding, private id: string) {
+    super(c, intDecoding);
+    this.id = id;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  path() {
+    return `/v2/deltas/txn/group/${this.id}`;
+  }
+}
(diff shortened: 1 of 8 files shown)


src/client/v2/algod/getLedgerStateDeltaForTransactionGroup.ts
```diff
@@ -0,0 +1,15 @@
+import JSONRequest from '../jsonrequest';
+import HTTPClient from '../../client';
+import IntDecoding from '../../../types/intDecoding';
+
+export default class GetLedgerStateDeltaForTransactionGroup extends JSONRequest {
```
algochoi, line 5: Is there no generated type for a `StateDelta` object?

algochoi (commented):
Looks good - wondering if ledger state delta objects are not generated below:

src/client/v2/algod/getLedgerStateDeltaForTransactionGroup.ts
algochoi: I forgot this was hand generated and not specified in the OAPI spec -- resolving

Eric-Warehime: Right, for now the StateDelta will be under-specified. Until we have something like state commitments that puts ledger objects behind consensus I don't feel comfortable maintaining delta objects and their internals for SDKs.

tests/cucumber/unit.tags
```diff
@@ -24,7 +24,9 @@
 @unit.responses.blocksummary
 @unit.responses.sync
```
algochoi, line 25: Consider Adding: `@unit.responses.statedelta`

algochoi (approved):
Suggestion to add statedelta response tag, otherwise LGTM

winder (approved):

PR #787, opened 2023-05-23
deps: Remove coveralls, mocha lcov

Eric-Warehime:
Remove coveralls and the mocha package used w/ coveralls.

We currently don't use it for coverage data--we aren't using any coverage packages as far as I can tell.

Ideally we can establish coverage metrics--potentially through the cucumber tests--but for now let's just remove these.


Eric-Warehime:
@algochoi Yeah I think `husky` is used in the git hooks. Some of the others I'm not sure about. Removing coveralls resolves a security warning stemming from the transitive dependency on `request` which is what originally led me to look into this.

algochoi (approved):
This looks fine - it was introduced [here](https://github.com/algorand/js-algorand-sdk/commit/cef8ef2659b5b6408e5498eaac97b470c4f9c3fe#diff-6d3cb368a70d68c76708298309adff01753f4d0873247afefbefeb9859e9439d) but I don't think it was being used then either 👍

Issue #771, opened 2023-04-26
Convert all client responses to typed objects


Eric-Warehime:
## Problem

Responses from the client are not automatically converted into the defined model types (for all endpoints).

We've done this for some endpoints (see https://github.com/algorand/js-algorand-sdk/blob/a89046acaf9502f496fca9d781bf33047c969c01/src/client/v2/algod/getApplicationBoxes.ts#L57-L60), but other endpoints, in order to avoid backwards incompatible changes, do not implement this (see https://github.com/algorand/js-algorand-sdk/blob/a89046acaf9502f496fca9d781bf33047c969c01/src/client/v2/algod/getApplicationByID.ts for example).

## Solution

Implement the `response.from_obj_for_encoding(body)` for all endpoints which return structured data.

## Dependencies

N/A

## Urgency

We should group this with any other breaking changes being made in the next major version bump.

jasonpaulos:
If we want to make typed response objects available sooner, we could consider introducing a new set of functions that can be used in place of `do`, e.g. `doTyped` or `execute`, which would return a typed response. We could then consider deprecating and removing `do` in v3 -- this would arguably be a safer way of swapping the behavior, instead instead of changing a function's return type, we'd be removing the function in favor of a different one.

Eric-Warehime:
An additional issue to consider w/ this mentioned in #533 is handling the conversion of the returned keys which include hyphens to the JS camel case keys.

PR #791, opened 2023-05-26
Adds typeVersions to package.json …

tristanmenzel:
Adds typeVersions to package.json so imports that don't target the root index.js file still have correct typings.

Currently having only the `types` property set means we only get typings when importing from file specified by the `main` property. Imports from sub-paths do not pick up correct typings.

eg.
image

Including this typeVersions section results in the imports being correctly typed.

image

An alternative solution would be to generate a full `exports` section then we could omit the `dist/esm` or `dist/cjs` section from the imports but this would be a breaking change for any existing usages and I'm hoping we can push this out as a patch release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,19 @@
     "crypto": false
   },
   "types": "dist/types/index.d.ts",
+  "typesVersions": {
+    "*": {
+      ".": [
+        "dist/types/index.d.ts"
+      ],
+      "dist/cjs/*": [
+        "dist/types/*"
+      ],
+      "dist/esm/*": [
+        "dist/types/*"
+      ]
+    }
+  },
   "files": [
     "dist/",
     "src/"


jasonpaulos:
Hi @tristanmenzel, thanks for making this PR. My only concern is that we want library users to only import the top-level `algosdk` package. It would be much harder for us to maintain and make guarantees about specific files in our repo.

Can you explain why you find it useful or necessary to import specific files?

tristanmenzel:
@jasonpaulos The main reasons is that the algod client currently has a mix of mapped and unmapped api responses. The relevant issue to go through and map all responses is [here](https://github.com/algorand/js-algorand-sdk/issues/771). Until this is done the easiest way to work around the issue is to import the models and manually call `from_obj_for_encoding` on the responses.  

eg. 
```ts
const status = NodeStatusResponse.from_obj_for_encoding(await algod.status().do())
```

Once issue 771 is resolved I don't think there's any need to do actual imports of the models but we would still need to do type only imports in order to be able to refer to the response types by name. 

eg.
```ts
import type { TealKeyValue } from 'algosdk/dist/types/client/v2/algod/models/types';
```

In this scenario it would probably be preferable to not have `/dist/types` in the path, and just import from `algosdk/client/v2/algod/models/types`. This could be achieved with a custom `exports` section in the package.json (requires "moduleResolution": "Node16" in your tsconfig.json)

```json
  "exports": {
    ".": {
      "types": "dist/types/index.d.ts",
      "require": "dist/cjs/index.js",
      "import": "dist/esm/index.js"
    },
    "./client/v2/algod/models/types": {
      "types": "./dist/types/client/v2/algod/models/types.d.ts",
      "require": "./dist/cjs/client/v2/algod/models/types.js",
      "import": "./dist/esm/client/v2/algod/models/types.js"
    },
  },
```
The great thing about the exports solution is you can selectively expose files which make sense (at least for clients which support it)

jasonpaulos:
Thanks for that detailed response, that was really helpful.

We do currently export the model types you mentioned under the `modelsv2` object. Docs here: https://algorand.github.io/js-algorand-sdk/modules/modelsv2.html

You should be able to achieve what you want by accessing this object, e.g.

```javascript
import { modelsv2 } from 'algosdk';

modelsv2.NodeStatusResponse.from_obj_for_encoding(...);
```

I realize it's not exactly the same thing, since the classes are namespaced under `modelsv2`, so please let us know if you find this to be an acceptable solution.

tristanmenzel:
Thanks @jasonpaulos, that does actually work for me.  I can just re-export the ones I'm using if I don't want to use the namespace. 

```ts
export const PendingTransactionResponse = modelsv2.PendingTransactionResponse
export type PendingTransactionResponse = modelsv2.PendingTransactionResponse
```

PR #794, opened 2023-06-06
fetch: Add Cloudflare Workers support (cross-fetch v4)

spencercap:
at the @thencc we're using the js algosdk in cloudflare workers' v8 environment for EXTREME server-side benefits. 

by simply updating to [`cross-fetch` v4](https://www.npmjs.com/package/cross-fetch/v/4.0.0-alpha.5?activeTab=versions) (currently in alpha), `algosdk` can work in [chrome v8](https://www.cloudflare.com/learning/serverless/glossary/what-is-chrome-v8/) in addition to the usual node + browser envs. this was just recently made available as discussed here: https://github.com/lquixada/cross-fetch/pull/163 

as a byproduct, this PR removes the `mode: "cors"` arg in `algosdk`'s fetch calls as it is not supported by the fetch api available within workers. may i ask what is the real use for this arg anyway?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "algo-msgpack-with-bigint": "^2.1.1",
         "buffer": "^6.0.3",
-        "cross-fetch": "^3.1.5",
+        "cross-fetch": "^4.0.0-alpha.5",
         "hi-base32": "^0.5.1",
         "js-sha256": "^0.9.0",
         "js-sha3": "^0.8.0",
@@ -1937,11 +1937,11 @@
       "dev": true
     },
     "node_modules/cross-fetch": {
-      "version": "3.1.5",
-      "resolved": "https://registry.npmjs.org/cross-fetch/-/cross-fetch-3.1.5.tgz",
-      "integrity": "sha512-lvb1SBsI0Z7GDwmuid+mU3kWVBwTVUbe7S0H52yaaAdQOXq2YktTCZdlAcNKFzE6QtRz0snpw9bNiPeOIkkQvw==",
+      "version": "4.0.0-alpha.5",
+      "resolved": "https://registry.npmjs.org/cross-fetch/-/cross-fetch-4.0.0-alpha.5.tgz",
+      "integrity": "sha512-JyOWYhayN5jX+vxmZC0w15BMUEVVmeHCodh5vfJQIFDF9jZZNfdQUFzZkviNtXMh+HwODYFxLnCOb7ZiztAdrw==",
       "dependencies": {
-        "node-fetch": "2.6.7"
+        "node-fetch": "^2.6.11"
       }
     },
     "node_modules/cross-spawn": {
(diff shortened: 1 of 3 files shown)


Eric-Warehime:
I'll take a look at this--I'm not really familiar with how the CORS stuff is currently implemented in the SDK. The CI is failing on unrelated things which will be fixed once we merge #793 I'll rerun once I merge that.

spencercap:
ok thanks ! yes, let me know what y'all think about removing that cors arg + that merge.

Eric-Warehime (approved):
Ok, I don't have a problem with this as is. We'll have to keep an eye on cross-fetch to see when we can just swap to v4 stable.

For the cors stuff it looks like it was committed [when we swapped from superagent to fetch](https://github.com/algorand/js-algorand-sdk/commit/2efd9c84b6567396051483df80548baf7efd0c7e), but the fetch library [doesn't provide support](https://github.com/node-fetch/node-fetch/blob/v2.6.7/README.md?plain=1#L396) for setting [the mode property](https://developer.mozilla.org/en-US/docs/Web/API/Request/mode) in requests, so it doesn't do anything.

jasonpaulos:
While the `cors` option has no affect for node, I'm under the impression that it's necessary for browsers. We need to carefully verify any changes here do not adversely affect browser usage.

Eric-Warehime:
@jasonpaulos 
> We need to carefully verify that any changes here do not adversely affect browser usage.

Agreed!

> While the cors option has no effect for node, I'm under the impression that it's necessary for browsers.

The node-fetch library which is being folyfilled by cross-fetch doesn't ever use the `mode` argument which is passed in. I've taken a look at the polyfill code I could find in the cross-fetch repo and I can't find anywhere that it's being used in there either. 

Is there some place you know of that I'm not finding w/r/t the creation of the request object in browsers?

jasonpaulos:
> The node-fetch library which is being folyfilled by cross-fetch doesn't ever use the `mode` argument which is passed in. I've taken a look at the polyfill code I could find in the cross-fetch repo and I can't find anywhere that it's being used in there either.
> 
> Is there some place you know of that I'm not finding w/r/t the creation of the request object in browsers?

The polyfills you mentioned should only ever be used in node. Our expectation is that the browsers we support are modern enough to supply their own native implementation of fetch -- no polyfill should be active for them. That's the environment that I think your analysis is missing.

In this case, the best way to test that environment is to access a webpage from a real host (i.e. not localhost or accessed as a local file) using the SDK, and having code which talks to an algod located in a different host. That's the situation in which CORS would apply.

Eric-Warehime:
> The polyfills you mentioned should only ever be used in node.

I guess that includes react-native as well--I know we don't support it right now.

Good point about the browsers. I'm unsure if our current browser tests will exercise cross origin support or not

youraerials:
Also, FWIW mode: "cors" is the default in the fetch api, you can safely omit it! https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

Eric-Warehime:
@spencercap The CI should pass if you rebase?

Also I'm happy to merge this as long as @jasonpaulos doesn't object.

Eric-Warehime:
It looks like the new CI failures are because we're bumping the golang version everywhere, but we haven't updated sandbox yet. I've got https://github.com/algorand/sandbox/pull/189 open to fix that and will re-run the CI once it's merged.

spencercap:
any estimate when this will be pushed to the @latest release?

PR #798, opened 2023-06-27
logging: Add rn warning and logging

Eric-Warehime:
Adds a log when we detect we're running in react-native which tells users they will need to polyfill crypto libraries.

The react-native project does not (and probably won't) provide any crypto functionality https://github.com/facebook/react-native/issues/5049 so users need to polyfill things like `crypto.getRandomValues` if they want to do address generation.

I've added a convenience function to determine if we're in react-native ( see https://github.com/facebook/react-native/commit/e966cd104110a980f343b794d85ac1b3afb2449e )

Also added `tslog` to provide logging cross platform and an exported function for users to adjust the log level if they don't want to see any logs.


src/logging.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * LOG is the SDK's base logger
+ */
+export const LOG = new Logger();
+
+/**
+ * setLogLevel adjusts the minimum level of logs which are output by the SDK's logger
+ */
+export function setLogLevel(level: number) {
```
tzaffi, line 11: Is this somehow being called implicitly somewhere?

tzaffi: Reading your PR description more carefully I see the purpose now. Perhaps we have enough static checks to see that this function is well formed. Regardless, a unit test can act as a sanity check.

Eric-Warehime: It's not being called anywhere, it's set that users can set the log level to turn off logs if they don't want to see a certain level.

winder (approved):
My only concern is whether or not we should document this behavior somehow. I guess the log is to make it self-documenting so I'm also OK with it as-is. I don't have any opinion on the logging library, but it seems like a sensible approach to providing this hint to the developers.

Blackglade (commented):
Looks good. Just wary of adding a logging specific package when logging isn't prevalent in the app to begin with. I think this is the only place (outside maybe some examples) that a log function is called. Maybe worth just switching to `console.warn()` until there's a specific need for more custom sdk-wide logging. _(unless there are plans in future PR's to introduce a lot more logging)_

Eric-Warehime:
@Blackglade Yeah I can change to a console log. My inclination is to introduce more logging in the future, but if it's not something that people want/need then yeah minimizing the dependencies is best.

Issue #759, opened 2023-04-05
Expose ATC transactions field as `protected`


barnjamin:
## Problem

Right now the ATC holds an array of TransactionWithSigner objects that is private

https://github.com/algorand/js-algorand-sdk/blob/develop/src/composer.ts#L120

This means to get access to the transactions, the developer must call `buildGroup` which changes the state of the ATC. 

It is possible to `clone` the ATC prior to calling `buildGroup` so that the state of the original ATC is unchanged but this is not obvious.

## Solution

Change `private` to `protected`

robdmoore:
Ben raised this after I provided some feedback to him about how ATC has a bunch of useful logic, but it's not very flexible for extension, hiding most of it away.

Agreed making protected isn't right, but I'd like to request the following two changes which would significantly improve the flexibility / extensibility:

* Return the transaction(s) / method call(s) that a call to addMethodCall creates (it may add multiple transactions, if they aren't returned then you have to do a `.clone().buildGroup()` to get the transactions every time you call it which is tedious and inefficient)
* Return the PendingTransactionInfo from execute (otherwise if you want that object you have to issue an additional algod call to get that data)

jasonpaulos:
Thanks for that insight @robdmoore. I just have a few questions:

> * Return the transaction(s) / method call(s) that a call to addMethodCall creates (it may add multiple transactions, if they aren't returned then you have to do a `.clone().buildGroup()` to get the transactions every time you call it which is tedious and inefficient)

This seems like a reasonable request. Can I ask why you're interested in seeing the transactions produced by addMethodCall though?

> * Return the PendingTransactionInfo from execute (otherwise if you want that object you have to issue an additional algod call to get that data)

We currently do return the PendingTransactionInfo for all method calls from execute. They're accessible in `ABIResult.txInfo`. Is this what you're looking for, or do you also want PendingTransactionInfo for non-method calls?

robdmoore:
Apologies for the delay in replying.

I want to see the transactions from `addMethodCall` because I have a wrapper that provides specific functionality that results in a call to `addMethodCall` and one of the return values from this method is the list of transaction(s) it generated. As an escape hatch and to allow for more advanced chaining, this method optionally takes an Atomic Transaction Composer so I can't just grab every single method in it since it may already have some methods.

[...]

PR #799, opened 2023-06-28
ATC: addMethodCall, addTransaction returns this

joe-p:
This simply changes the return type of `addTransaction` and `addMethodCall` from `void` to `AtomicTransactionCompose` so these methods return the ATC. 

The goal is to allow chaining of methods when using the ATC rather than having to instantiate an object and call the methods on that object each time. 

For example:
```typescript
  const atc = new algosdk.AtomicTransactionComposer();
  atc.addTransaction();
  atc.addMethodCall();
  atc.execute();
```

vs

```typescript
  new algosdk.AtomicTransactionComposer()
    .addTransaction()
    .addMethodCall()
    .execute();
```

No major reason other than a nice minor QOL improvement

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -169,7 +169,9 @@ export class AtomicTransactionComposer {
    * An error will be thrown if the transaction has a nonzero group ID, the composer's status is
    * not BUILDING, or if adding this transaction causes the current group to exceed MAX_GROUP_SIZE.
    */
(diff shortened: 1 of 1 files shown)


jasonpaulos:
Note that this is in conflict with @robdmoore's desire to have `addMethodCall()` return the newly created transaction: https://github.com/algorand/js-algorand-sdk/issues/759#issuecomment-1505550812

PR #800, opened 2023-07-06
DRAFT: Remove buffer usage in favor or Uint8Array and Dataview

algochoi:


src/utils/utils.ts
```diff
@@ -122,3 +123,52 @@ export function isNode() {
...
+/**
+ * Convert a base64 string to a Uint8Array for Node.js and browser environments.
+ * @returns A Uint8Array
+ */
+export function base64ToBytes(base64String: string) {
+  if (isNode()) {
+    return Buffer.from(base64String, 'base64');
+  }
+  // eslint-disable-next-line no-undef
+  return new Uint8Array(new TextEncoder().encode(btoa(base64String)));
```
Eric-Warehime, line 136: These are temporary conversion functions or permanent?

src/utils/utils.ts
```diff
@@ -122,3 +123,52 @@ export function isNode() {
     typeof process.versions.node !== 'undefined'
   );
 }
+
+/**
+ * Convert a base64 string to a Uint8Array for Node.js and browser environments.
+ * @returns A Uint8Array
+ */
+export function base64ToBytes(base64String: string) {
```
jasonpaulos, line 131: minor: could you explicitly annotate these functions with return values as well?

While Typescript can infer them, I'd find the extra clarify helpful, and it ensures the return type won't accidentally change if we update the function body.

src/utils/utils.ts
```diff
@@ -122,3 +123,52 @@ export function isNode() {
...
+  if (isNode()) {
+    return new Uint8Array(Buffer.from(base64String, 'base64'));
+  }
+  // eslint-disable-next-line no-undef
```
jasonpaulos, line 135:
```suggestion
  /* eslint-env browser */
```
FYI the reason we see an error here by default is because we specify our global eslint environment as [`shared-node-browser`](https://github.com/algorand/js-algorand-sdk/blob/72653522835914c451357db9c1f98eb79ddbdf2a/.eslintrc.js#L4) -- only the definitions common to both environments are allowed by default.

Since this code branch should only get executed in the browser, we can make eslint aware of this with this comment, which is safer than ignoring all `undef` errors.

tests/2.Encoding.js
```diff
@@ -538,4 +552,18 @@ describe('encoding', () => {
       }
     });
   });
+
+  describe('Base64 decoding utilities', () => {
+    it('should decode and encode Base64 roundtrip', () => {
```
jasonpaulos, line 557: I have concerns about `TextEncoder`/`TextDecoder` and `btoa`/`atob` working with non-ASCII characters. See [The "Unicode Problem"](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem)

Additional tests here with special Unicode characters (e.g. "a Ā 𐀀 文 🦄" from the link above), as well as 0s and other low-value characters would help prove these conversion functions work as intended.

algochoi: Good point! Added new tests and changed browser implementation to mirror the one in the MDN docs

Eric-Warehime (approved):
LGTM, needs rebase it looks like?

tests/4.Utils.ts
```diff
@@ -34,6 +34,56 @@ describe('utils', () => {
...
     });
   });
+
+  describe('Base64 decoding utilities', () => {
+    it('should decode and encode Base64 roundtrip', () => {
+      const testCases = [
+        ['Hello, Algorand!', 'SGVsbG8sIEFsZ29yYW5kIQ=='],
+        ['a Ā 𐀀 文 🦄', 'YSDEgCDwkICAIOaWhyDwn6aE'],
+        ['(╯°□°）``` ┻━┻ 00\\', 'KOKVr8Kw4pahwrDvvIlgYGAg4pS74pSB4pS7IDAwXA=='],
+      ];
```
jasonpaulos, line 44: I added this additional case, and unfortunately there was an error:

`['\x00\x01\x02\x03\x04...\xFE\xFF', 'AAECAwQuLi7+/w==']`

```
      AssertionError [ERR_ASSERTION]: Incorrect encoding of ...þÿ; got ...��
      + expected - actual

      -...��
      +...þÿ
      
      at Context.<anonymous> (tests/4.Utils.ts:48:16)
      at processImmediate (node:internal/timers:476:21)
```

Could you take a look at this?

algochoi: You're right, it does not work for non UTF-8 strings -- the following base64 string (which is different from yours) works, so I guess the encoder/decoder provides a best attempt at it:
`['\x00\x01\x02\x03\x04...\xFE\xFF', 'AAECAwQuLi7DvsO/']`

Should we use an external dependency like [base64-js](https://github.com/beatgammit/base64-js) for arbitrary binary data? It looks like we didn't do this for non UTF-8 strings to begin with.

jasonpaulos: If I use the exact conversion functions from https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem (copied below), I find it is able to properly encode and decode the base64 string `AAECAwQuLi7+/w==` to the Uint8Array `[0, 1, 2, 3, 4, 46, 46, 46, 254, 255]`.

So I am confident base64 <=> Uint8Array conversion can work with this method.

Maybe something's going wrong with the Uint8Array <=> string conversion? I will try to look at that more closely soon.

```javascript
function base64ToBytes(base64) {
  const binString = atob(base64);
  return Uint8Array.from(binString, (m) => m.codePointAt(0));
}

function bytesToBase64(bytes) {
  const binString = Array.from(bytes, (x) => String.fromCodePoint(x)).join("");
  return btoa(binString);
}
```

If you find this implementation works equally well

examples/app.ts
```diff
@@ -4,9 +4,9 @@
 /* eslint-disable no-console */
 import fs from 'fs';
 import path from 'path';
-import { Buffer } from 'buffer';
 import { getLocalAlgodClient, getLocalAccounts, compileProgram } from './utils';
 import algosdk from '../src';
+import { base64ToBytes, base64ToString } from '../src/utils/utils';
```
jasonpaulos, line 9: Since you are importing these in the examples, that shows these conversion functions are going to be very useful to consumers as well.

For that reason, I think it would be a good idea to export them to be easily used by consumers. We already have the `src/encoding` directory where we put encoding functions which are later exported (see `src/main.ts`), and I think these would fit in there really well.

src/utils/utils.ts
```diff
@@ -122,3 +123,56 @@ export function isNode() {
...
+  /* eslint-env browser */
+  const binString = atob(base64String);
+  return Uint8Array.from(binString, (m) => m.codePointAt(0));
+}
+
+/**
+ * Decode a base64 string for Node.js and browser environments.
+ * @returns A decoded string
+ */
+export function base64ToString(base64String: string): string {
```
jasonpaulos, line 144: I'm wondering if it would be more modular to offer a separate set bytes <=> string conversion functions. For instance, if I had a hex string I wanted to convert to a normal string, right now I'm not able to do that.

These might help us pinpoint the current issue we're debugging as well.

tests/4.Utils.ts
algochoi: Yes, I used the same implementation as that one -- the problem only exists when trying to go from Uint8Array with non UTF-8 characters to a (UTF-8) string and back into something else.

If we're satisfied with just checking that the base64 (with non UTF-8 bytes) can go to binary data, then we're good and I can update the tests for this case.

algochoi: I added a separate test case for this in the latest commit as an example: https://github.com/algorand/js-algorand-sdk/pull/800/files#diff-9edc36253646a135dddbc24bad91040c7e13cda76d51a79e2c8737d082a85de9R65

examples/app.ts
algochoi: I'm on board with this - I separated it out into `binarydata.ts` under `encoding/` and exported it in `main.ts`

tests/4.Utils.ts
jasonpaulos: After more research, I've concluded that since Javascript strings are limited to UTF-8, it makes sense that they cannot hold those invalid UTF-8 characters (e.g. `\xff`).

As long as the new `TextDecoder` code path produces the same output as the `Buffer` codepath does for these invalid characters, this seems like an ok result to me. Could you add a test that checks this?

examples/app.ts
```diff
@@ -4,9 +4,9 @@
 /* eslint-disable no-console */
 import fs from 'fs';
 import path from 'path';
-import { Buffer } from 'buffer';
 import { getLocalAlgodClient, getLocalAccounts, compileProgram } from './utils';
 import algosdk from '../src';
+import { base64ToBytes, base64ToString } from '../src/encoding/binarydata';
```
jasonpaulos, line 9: Could you remove this import and instead reference the functions like `algosdk.base64ToBytes`? That's how we prefer consumers to import definitions

src/utils/utils.ts
jasonpaulos: Feel free to downgrade this to a nit -- I don't feel as strongly anymore, and we can easily add more util functions if we find they'd be useful.

jasonpaulos (commented):
Just some final thoughts -- once the import issue in the examples is fixed and the new text comparison test is added, I'll approve

examples/app.ts
algochoi: Good catch, changed the functions to prefix with `algosdk`

tests/4.Utils.ts
algochoi: Thanks for the double check, I added a test for this here (with only UTF-8 characters): https://github.com/algorand/js-algorand-sdk/pull/800/files#diff-03788da9bcb72be6b53a86f12c45248552ca8732d28938fc18d6bdebdbf64ccfR604

It should run both code paths (`Buffer` in node and `TextDecoder` in browser) in CI.

examples/app.ts
jasonpaulos: The other example files need to be updated as well

tests/4.Utils.ts
algochoi: I added `'/v8='` to the roundtrip test case that checks for `base64ToString` conversions - let me know if it resolves this: [92333c2](https://github.com/algorand/js-algorand-sdk/pull/800/commits/92333c25147cc55a3ca15c0bfffd714c77d48b93)

jasonpaulos (approved):

Issue #734, opened 2023-01-14
Stop using `Buffer` in browser code


jasonpaulos:
## Summary

We use the npm [buffer](https://www.npmjs.com/package/buffer) package to allow our library to use node's `Buffer` class in the browser.

I think we should stop doing this for a few reasons:
* [buffer](https://www.npmjs.com/package/buffer) does not appear to be maintained anymore, as it hasn't been updated in ~2 years and has numerous issue and PRs open.
* Webpack v4 projects provide an even older version of [buffer](https://www.npmjs.com/package/buffer), and we're unable to do anything about it. This is pretty concerning, as we don't test with this older version.
* Most of our `Buffer` usage seems to be encoding and decoding base64. We should be able to do this in the browser without `Buffer`.

## Suggestions

Instead of relying on a browser `Buffer` implementation, we could make better use of `Uint8Array` and `Dataview`, which are supported in both node and browser. Some things will still require `Buffer` usage in node (like base64 encoding), but we can investigate ways to have an alternative code path for browsers.

There are a few public APIs which reference `Buffer`, so changing these would likely have to wait for a major release of the library.

paulmillr:
https://github.com/algorand/js-algorand-sdk/issues/742 doesn't use buffers

algoanne:
> Some things will still require Buffer usage in node (like base64 encoding), but we can investigate ways to have an alternative code path for browsers.

@jasonpaulos why does base64 encoding require it?

jasonpaulos:
@algoanne require is a bit of a strong word, but using `Buffer` for base64 encoding is likely the easiest/most efficient option for node, so that's why I suggested continuing to use it there. It is however not a strict requirement.

PR #801, opened 2023-07-19
Remove `Buffer` Usage

algochoi:
Follow up PR to remove `Buffer` usage in JS SDK:
* Isolate `Buffer` usage to node environment
* Fix mocha and cucumber tests to use `Uint8Array`
* Generator changes: https://github.com/algorand/generator/pull/65

Some extra changes:
* Increase indexer sleep time to fix flaky `asa` example
* Re-arrange some imports 

Closes https://github.com/algorand/js-algorand-sdk/issues/734


examples/asa.ts
```diff
@@ -49,7 +49,7 @@ async function main() {
   console.log(`Asset Params: ${JSON.stringify(assetInfo.params)}`);
   // example: ASSET_INFO
 
-  await new Promise((f) => setTimeout(f, 2000)); // sleep to ensure indexer is caught up
+  await new Promise((f) => setTimeout(f, 5000)); // sleep to ensure indexer is caught up
```
algochoi, line 52: This looks frequently flaky on CI, but I haven't seen it fail locally 👀

src/encoding/binarydata.ts
```diff
@@ -53,3 +64,22 @@ export function bytesToHex(byteArray: Uint8Array): string {
...
+  if (isNode()) {
+    return Buffer.from(hexString, 'hex');
+  }
+  let hex = hexString;
+  if (hexString.length % 2 !== 0) {
+    hex = hexString.padStart(1, '0');
+  }
+  const byteArray = new Uint8Array(hex.length / 2);
+  for (let i = 0, j = 0; i < hex.length / 2; i++, j += 2) {
+    byteArray[i] = parseInt(hex.slice(j, j + 2), 16);
```
Eric-Warehime, line 82:
```suggestion
  for (let i = 0; i < hex.length / 2; i++) {
    byteArray[i] = parseInt(hex.slice(2*i, 2*i + 2), 16);
```

Eric-Warehime: Either way seems fine--this way you don't need double iter vars though which feels cleaner to me.

src/client/v2/algod/models/types.ts
```diff
@@ -3,10 +3,10 @@
  */
```
Eric-Warehime, line 3: Needs https://github.com/algorand/generator/pull/65/ merged.

src/transaction.ts
```diff
@@ -1261,7 +1259,8 @@ export class Transaction implements TransactionStorageStructure {
       forPrinting.reKeyTo = address.encodeAddress(
         (forPrinting.reKeyTo as Address).publicKey
       );
-    forPrinting.genesisHash = forPrinting.genesisHash.toString('base64');
+    if (typeof forPrinting.genesisHash !== 'string')
+      forPrinting.genesisHash = bytesToBase64(forPrinting.genesisHash); // TODO: double check this
```
Eric-Warehime, line 1263: done?

tests/cucumber/steps/index.js
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable no-console,global-require,no-loop-func,func-names */
 const assert = require('assert');
-const { Buffer } = require('buffer');
```
jasonpaulos, line 3: This file only ever gets executed in node to set up the cucumber tests, so it would be perfectly acceptable to still use Buffer here

jasonpaulos (commented):
I think we'd also want to remove the dependency on the 3rd party `buffer` package to consider #734 complete. Do you plan to do that in this PR?

tests/cucumber/steps/index.js
algochoi: Gotcha, reverted this back in [50e03cf](https://github.com/algorand/js-algorand-sdk/pull/801/commits/50e03cf358c119529a9bb4019962d9cabb4fe723)

algochoi: Also good point on removing the dependency -- I removed the browser buffer package and removed the lint rule around `Buffer` in [e7aa3d3](https://github.com/algorand/js-algorand-sdk/pull/801/commits/e7aa3d3f3694a82a67f69c837777198777ca70ae)

src/transaction.ts
algochoi: good catch, will remove this

Eric-Warehime (approved):

tests/cucumber/steps/steps.js
```diff
@@ -41,6 +40,10 @@ async function loadResource(res) {
   });
 }
 
+async function loadResourceAsJson(res) {
+  return JSON.parse(await loadResource(res));
```
jasonpaulos, line 44: `loadResource` returns a `Buffer`, I think you'll need to convert to a string before JSON parsing

jasonpaulos (commented):
Will approve once the `loadResourceAsJson` issue is resolved

tests/cucumber/steps/steps.js
algochoi: Interesting, I didn't see this originally converted to a string, but I added the conversion just in case [ee3e2e1](https://github.com/algorand/js-algorand-sdk/pull/801/commits/ee3e2e12b1260fad5a966352d16dfb256f0906e2)

jasonpaulos (approved):

Issue #144, opened 2020-05-05
Document or don't rename transaction fields


jeapostrophe:
The Algorand document lists names for the fields --- https://developer.algorand.org/docs/reference/transactions/ --- and the JS SDK (maybe the other's too, but I don't know) doesn't uses these names.

For example, the JS SDK uses `firstRound` rather than `FirstValid` (field name) or `fv` (codec name). Basically every field has a different name.

There's no documentation that I can find anywhere that lists what the actual names the JS SDK uses. Instead, I just have to always have the source code open (https://github.com/algorand/js-algorand-sdk/blob/develop/src/transaction.ts#L70)

This is really disruptive and difficult for developing.

algorotem:
Hi @jeapostrophe Thanks for the comment and apologies for the delayed response. 
The fields names in the doc you linked to are referring to the names of the fields in the actual msgpack encoding of the transaction and not in the SDKs. That being said, it's a great suggestion and we will get out a conversion table as soon as possible to make it easier.

Eric-Warehime:
I think it makes a lot of sense to change the naming here to be consistent with our naming spec for Transactions. But doing so would mean a breaking change. I'm adding this issue to https://github.com/algorand/js-algorand-sdk/issues/654 to track it whenever we want to make a major version bump.

algoanne:
I went through the list, here are names that are confusingly different (there are others that have some variation but semantically the same):

from - Sender 
to - Receiver or AssetReceiver 
firstRound - firstValid
lastRound - lastValid
freezeState - AssetFrozen
appIndex - Application ID 
assetIndex - ConfigAsset (codec: caid), XferAsset (codec: xaid), FreezeAsset (codec: faid)
assetRevocationTarget - Sender when it's clawback? 

It's not totally clear to me whether it's worth changing some of these now (maybe it was 3 years ago when this issue was opened). We are about to do a major version bump so if we want to change them, now would be a good time. But this will likely cause a good amount of churn for people using this SDK when they upgrade to version 3.

algochoi:
Closed by #804 -- we have closed the gap on naming consistencies, and the changes will go out in a breaking version bump

PR #804, opened 2023-08-01
Align transaction fields to transaction reference spec

algochoi:
This PR contains breaking changes in transaction fields that align more closely to the [transaction reference spec](https://developer.algorand.org/docs/get-details/transactions/transactions/).

Changes:
* from -> sender
* to -> receiver
* firstRound -> firstValid
* lastRound -> lastValid
* freezeState -> assetFrozen
* assetRevocationTarget -> assetSender
* reKeyTo -> rekeyTo

https://github.com/algorand/js-algorand-sdk/issues/144

### DIFF
--- a/src/types/transactions/base.ts
+++ b/src/types/transactions/base.ts
@@ -311,7 +311,7 @@ export interface TransactionParams {
   /**
    * true if freezeTarget should be frozen, false if freezeTarget should be allowed to transact
    */
-  freezeState: boolean;
+  assetFrozen: boolean;
 
   /**
    * String representation of Algorand address – if provided, and if "from" is
--- a/src/types/transactions/base.ts
+++ b/src/types/transactions/base.ts
@@ -162,12 +162,12 @@ export interface TransactionParams {
   /**
    * String representation of Algorand address of sender
    */
-  from: string;
+  sender: string;
 
   /**
    * String representation of Algorand address of recipient
    */
-  to: string;
+  receiver: string;
 
   /**
    * Integer fee per byte, in microAlgos. For a flat fee, set flatFee to true
@@ -314,8 +314,8 @@ export interface TransactionParams {
   assetFrozen: boolean;
 
   /**
-   * String representation of Algorand address – if provided, and if "from" is
-   * the asset's revocation manager, then deduct from "revocationTarget" rather than "from"
+   * String representation of Algorand address – if provided, and if "sender" is
+   * the asset's revocation manager, then deduct from "revocationTarget" rather than "sender"
    */
   assetRevocationTarget?: string;
 
--- a/src/types/transactions/base.ts
+++ b/src/types/transactions/base.ts
@@ -112,12 +112,12 @@ export interface SuggestedParams {
   /**
    * First protocol round on which this txn is valid
    */
-  firstRound: number;
+  firstValid: number;
 
   /**
    * Last protocol round on which this txn is valid
    */
-  lastRound: number;
+  lastValid: number;
 
   /**
    * Specifies genesis ID of network in use
@@ -182,12 +182,12 @@ export interface TransactionParams {
   /**
    * Integer first protocol round on which this txn is valid
    */
-  firstRound: number;
+  firstValid: number;
 
   /**
    * Integer last protocol round on which this txn is valid
    */
-  lastRound: number;
+  lastValid: number;
 
   /**
    * Arbitrary data for sender to store
--- a/src/types/transactions/base.ts
+++ b/src/types/transactions/base.ts
@@ -399,7 +399,7 @@ export interface TransactionParams {
   /**
    * String representation of the Algorand address that will be used to authorize all future transactions
    */
-  reKeyTo?: string;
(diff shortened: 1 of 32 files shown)


algochoi:
@algoanne We have two places where we use `application-index` in our APIs: https://developer.algorand.org/docs/rest-apis/algod/#pendingtransactionresponse but the most recent ones refer to it as `application-id`. Do we want to make this change here for app/assets?

algoanne:
@algochoi it seems ID is more common. Disappointing that we're not fully consistent at the API level. I really don't feel strongly one way or the other, they both make me sad. I leave it to you.

algochoi:
They also make me sad, and I would like to consolidate everything to the newer convention of `IDs`. 

However, we also have places in the API that use `Index` such as Assets: https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/algod.oas2.json#L2863
Because of these inconsistencies that will also affect how the SDK types will be generated, e.g. https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/v2/algod/models/types.ts#L1049, I'm inclined to keep everything as-is to be consistent with the API spec. I think changing these index-related conventions should be following a breaking change in the API spec as well.

src/types/transactions/base.ts
```diff
@@ -182,12 +182,12 @@ export interface TransactionParams {
   /**
    * Integer first protocol round on which this txn is valid
    */
-  firstRound: number;
+  firstValid: number;
```
jasonpaulos, line 185: Can you also rename the `reKeyTo` field to `rekeyTo`?

algochoi: good point -- [1258d1a](https://github.com/algorand/js-algorand-sdk/pull/804/commits/1258d1a6864851b6a5816de5bd3428e52171fe4c)

jasonpaulos:
> They also make me sad, and I would like to consolidate everything to the newer convention of `IDs`.
> 
> However, we also have places in the API that use [`Index` such as Assets](https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/algod.oas2.json#L2863). Because of these inconsistencies that will also affect how the SDK types will be generated, [e.g. here](https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/v2/algod/models/types.ts#L1049), I'm inclined to keep everything as-is to be consistent with the API spec. I think changing these index-related conventions should be following a breaking change in the API spec as well.

I have a hard time believing that we'll want to break the REST APIs anytime soon with an index => ID change, so it's worth thinking about what we can do short of that solving the problem.

We do have some cases where the REST API uses ID instead of index, for example [here](https://github.com/algorand/js-algorand-sdk/blob/4e1073d6a15b443a73dedf42b025173b397624b0/src/client/v2/algod/models/types.ts#L1113).

One option we could consider is modifying the generator repo to rename certain fields from index => ID. Then we could use ID everywhere, including in the SDK defined classes.

algochoi:
> I have a hard time believing that we'll want to break the REST APIs anytime soon with an index => ID change, so it's worth thinking about what we can do short of that solving the problem.

I agree as well -- I wanted to skirt around this issue for now because it seemed like the most onerous change (requires some custom overrides in generator, and then it may affect the three SDKs we generate for in order to keep field names consistent with each other, plus any overhead that these changes might have in the tests/examples/dev docs). On the other hand, the current changes are pretty isolated to the JS SDK and are easier wins and has had enough notice to dev-rel.

algoanne:
> I agree as well -- I wanted to skirt around this issue for now because it seemed like the most onerous change (requires some custom overrides in generator, and then it may affect the three SDKs we generate for in order to keep field names consistent with each other, plus any overhead that these changes might have in the tests/examples/dev docs). On the other hand, the current changes are pretty isolated to the JS SDK and are easier wins and has had enough notice to dev-rel.

I think I support breaking out the index -> ID changes into a separate PR, since it potentially has broader scope (across all the SDKs). Although we have not been explicitly planning a 3.0 for the other SDKs, it's certainly on the table. It would be valuable to bring this same consistency to all the SDKs. We can discuss. 
Do the other SDKs have consistent naming with the API in the other fields, e.g. lastValid etc. ?

algochoi:
> Do the other SDKs have consistent naming with the API in the other fields, e.g. lastValid etc. ?

I have only checked the obvious ones listed above (sender, lastValid), but yes. The go-sdk closely follows go-algorand conventions, Java is pretty similar, and Python has some minor variations in naming but still makes sense to me. 

The generated SDKs (Go, Java) are also consistent with the API spec, which has some weird naming like using `index` for Assets but `asset-id` for AssetHoldings. So I think changing these would probably warrant a change in all the SDKs. I haven't thought about potential downstream impacts here (e.g. Swift, C# SDKs...).

winder:
I'm fine with ignoring the ID/Index issue, but we also have a lot of flexibility in the generator with applying things to a single SDK

Eric-Warehime (approved):

winder (approved):

bbroder-algo (approved):
s-tier

Issue #414, opened 2021-08-11
Remove function removeFalsyOrEmpty


barnjamin:
### Subject of the issue

I'm trying to lookup holders of a given asset where they hold >0 tokens.

When I call `client.lookupAssetBalances(asset_id).currencyGreaterThan(0).do()` the resultant request has no `currency-greater-than` GET parameter, but setting it to 1 does produce the correct request URL.

I believe this is due to the step where it removes things that look falsy:
https://github.com/algorand/js-algorand-sdk/blob/9afe012d48817978b3f49131dff86ccdb8c8d423/src/client/client.ts#L53-L61

I'm not sure what the correct fix is and I've got a temporary work around.

Lumene98:
@chaihoang Do you have updates on this?

chaihoang:
@jasonpaulos

jasonpaulos:
This is definitely a bug that we'd like to fix, but in the meantime community members have noticed that a workaround is to pass 0 as a string, i.e. `"0"`.

achidlow:
Are there any end-points in the API where it makes sense to filter out a query parameter that has been set to `0`? 

If not, can we just exclude that in `removeFalsyOrEmpty` like so:
`if (!obj[key] || obj[key].length === 0)`
becomes
`if ((!obj[key] && obj[key] !== 0) || obj[key].length === 0)`
?

d13co:
I hit this too - the `'0'` workaround does work but this should ideally be fixed, given how old this issue is. 

If you think excluding other falsy values is going to be problematic then at least make an exception for Number(0) in that check? "Holders that have at least one micro-unit" is a fairly common use case that everyone trips over currently.

shiqizng:
It may be possible to remove `removeFalsyOrEmpty`.  We need to fix errors like this after removing this function.  
```
✖ Then expect the path used to be "/v2/blocks/3?header-only=true" # tests/cucumber/steps/index.js:513
         AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
         + actual - expected

         + '/v2/blocks/3?accept=application%2Fjson'
         - '/v2/blocks/3?header-only=true'
                         ^
             + expected - actual

             -/v2/blocks/3?accept=application%2Fjson
             +/v2/blocks/3?header-only=true
```

PR #807, opened 2023-08-07
bug-fix: include currency-greater-than param for 0 value

### LINKED ISSUE #806: Copy of Valid GET parameter of 0 filtered out
### Subject of the issue

I'm trying to lookup holders of a given asset where they hold >0 tokens.

When I call `client.lookupAssetBalances(asset_id).currencyGreaterThan(0).do()` the resultant request has no `currency-greater-than` GET parameter, but setting it to 1 does produce the correct request URL.

I believe this is due to the step where it removes things that look falsy:
https://github.com/algorand/js-algorand-sdk/blob/9afe012d48817978b3f49131dff86ccdb8c8d423/src/client/client.ts#L53-L61

I'm not sure what the correct fix is and I've got a temporary work around.

### LINKED ISSUE #810: Enable 0 value for currency-greater-than
## Problem
When making the call `client.lookupAssetBalances(asset_id).currencyGreaterThan(0).do()` the request has no `currency-greater-than`

## Solution
Until #414 is resolved, treat 0 as '0' to get around this issue.

shiqizng:
This PR fixes #810

### DIFF
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="shiqi/pathupdates"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="shiqi/pathupdates"
+SDK_TESTING_BRANCH="V2"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/src/client/v2/indexer/lookupAccountTransactions.ts
+++ b/src/client/v2/indexer/lookupAccountTransactions.ts
@@ -305,6 +305,7 @@ export default class LookupAccountTransactions extends JSONRequest {
    * @category query
    */
   currencyGreaterThan(greater: number) {
+    // We convert the following to a string for now to correctly include zero values in request parameters.
     this.query['currency-greater-than'] = greater.toString();
     return this;
   }
--- a/src/client/v2/indexer/lookupAccountTransactions.ts
+++ b/src/client/v2/indexer/lookupAccountTransactions.ts
@@ -307,8 +307,7 @@ export default class LookupAccountTransactions extends JSONRequest {
    * @category query
    */
   currencyGreaterThan(greater: number) {
-    if (greater === 0) this.query['currency-greater-than'] = '0';
-    else this.query['currency-greater-than'] = greater;
+    this.query['currency-greater-than'] = greater.toString();
     return this;
   }
 
--- a/src/client/v2/indexer/lookupAccountTransactions.ts
+++ b/src/client/v2/indexer/lookupAccountTransactions.ts
@@ -300,8 +300,6 @@ export default class LookupAccountTransactions extends JSONRequest {
    *        .currencyGreaterThan(minBalance - 1)
    *        .do();
    * ```
-   * @remarks
-   * If you are looking for transactions with the currency amount greater than 0, simply construct the query without `currencyGreaterThan` because it doesn't accept `-1`, and passing the `0` `currency-greater-than` value would exclude transactions with a 0 amount.
    *
    * @param greater
    * @category query
--- a/src/client/v2/indexer/lookupAccountTransactions.ts
+++ b/src/client/v2/indexer/lookupAccountTransactions.ts
@@ -307,7 +307,8 @@ export default class LookupAccountTransactions extends JSONRequest {
    * @category query
    */
   currencyGreaterThan(greater: number) {
-    this.query['currency-greater-than'] = greater;
+    if (greater === 0) this.query['currency-greater-than'] = '0';
+    else this.query['currency-greater-than'] = greater;
     return this;
   }
 
--- a/src/client/v2/indexer/searchForTransactions.ts
+++ b/src/client/v2/indexer/searchForTransactions.ts
@@ -398,8 +398,7 @@ export default class SearchForTransactions extends JSONRequest {
    * @category query
    */
   currencyGreaterThan(greater: number) {
-    if (greater === 0) this.query['currency-greater-than'] = '0';
-    else this.query['currency-greater-than'] = greater;
+    this.query['currency-greater-than'] = greater.toString();
     return this;
   }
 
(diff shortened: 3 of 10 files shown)


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="shiqi/pathupdates"
```
shiqizng, line 3: TODO: update branch to V2 before merge.

tests/cucumber/steps/steps.js
```diff
@@ -1981,9 +1981,9 @@ module.exports = function getSteps(options) {
     ) {
       await this.indexerClient
         .lookupAssetBalances(index)
-        .limit(limit)
         .currencyGreaterThan(currencyGreater)
         .currencyLessThan(currencyLesser)
+        .limit(limit)
```
shiqizng, line 1986: params are reordered in this file to match the expected order.

algochoi (commented):
This looks fine, but I don't think it completely closes #414 -- it still uses the `'0'` string workaround, and I wonder if we can just change the behavior of `removeFalsyOrEmpty` instead: 
https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/client.ts#L26

I'm not sure if this would affect the messagepack serialization behavior though.

shiqizng:
> This looks fine, but I don't think it completely closes #414 -- it still uses the `'0'` string workaround, and I wonder if we can just change the behavior of `removeFalsyOrEmpty` instead: https://github.com/algorand/js-algorand-sdk/blob/develop/src/client/client.ts#L26
> 
> I'm not sure if this would affect the messagepack serialization behavior though.

I don't see a way to change removeFalsyOrEmpty so it allows `0` value without adding a special case for key `currency-greater-than`. 

@algorand/lamprey  what do others think about this? should we use a string work around or add a special handling in `removeFalsyOrEmpty` for individual fields?

algochoi:
> I don't see a way to change removeFalsyOrEmpty so it allows 0 value without adding a special case for key currency-greater-than.

I also don't see a very easy way to fix #414  without affecting serialization and breaking all our current tests. 

It seems potentially ugly, but I'd actually be in favor of what you suggested (having special cases). I think it's more obvious and self-documenting that this is a weird workaround in the JS SDK. As is, I don't think it's obvious to devs why we need to pass `'0'`s around without looking at this open issue (414), and especially if we have future client parameters that may require 0 as a valid parameter.

jasonpaulos:
It is possible we could stop running queries through the `removeFalsyOrEmpty` function? I'm not sure what benefits we even get from this. I'm pretty sure Python does not do something similar, and it certainly works.

shiqizng:
> It is possible we could stop running queries through the `removeFalsyOrEmpty` function? I'm not sure what benefits we even get from this. I'm pretty sure Python does not do something similar, and it certainly works.

If I remove this function, then the path seem to include `accept=application%2Fjson` which is causing the tests to fail. 
```
✖ Then expect the path used to be "/v2/blocks/3?header-only=true" # tests/cucumber/steps/index.js:513
         AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
         + actual - expected

         + '/v2/blocks/3?accept=application%2Fjson'
         - '/v2/blocks/3?header-only=true'
                         ^
             + expected - actual

             -/v2/blocks/3?accept=application%2Fjson
             +/v2/blocks/3?header-only=true
```

jasonpaulos:
That's very strange, I have no idea where it's getting `accept=application%2Fjson` from... 🤔

src/client/v2/indexer/lookupAccountTransactions.ts
```diff
@@ -307,7 +307,8 @@ export default class LookupAccountTransactions extends JSONRequest {
    * @category query
    */
   currencyGreaterThan(greater: number) {
-    this.query['currency-greater-than'] = greater;
+    if (greater === 0) this.query['currency-greater-than'] = '0';
+    else this.query['currency-greater-than'] = greater;
```
winder, line 311: What do you think about avoiding the condition and doing something like this?
```suggestion
    this.query['currency-greater-than'] = '' + greater;
```

or

```suggestion
    this.query['currency-greater-than'] = greater.toString();
```

src/client/v2/indexer/searchForTransactions.ts
```diff
@@ -432,4 +368,68 @@ export default class SearchForTransactions extends JSONRequest {
...
+   * const assetID = 163650;
+   * const minBalance = 300000;
+   * const txns = await indexerClient
+   *        .searchForTransactions()
+   *        .assetID(assetID)
+   *        .currencyGreaterThan(minBalance - 1)
+   *        .do();
+   * ```
+   * @remarks
+   * If you are looking for transactions with the currency amount greater than 0, simply construct the query without `currencyGreaterThan` because it doesn't accept `-1`, and passing the `0` `currency-greater-than` value would exclude transactions with a 0 amount.
```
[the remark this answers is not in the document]
algochoi: This seems still true, but the original wording is a bit confusing and verbose. I'd be okay with deleting this remark anyway.

bbroder-algo:
I think I would want to know what falsy values or values with length zero we are removing from queries, and whether it can be removed or replaced with alternate logic.  my hunch is the `accept=application%2Fjson` situation can be resolved.

shiqizng:
> I think I would want to know what falsy values or values with length zero we are removing from queries, and whether it can be removed or replaced with alternate logic. my hunch is the `accept=application%2Fjson` situation can be resolved.

This is to be addressed in #414

algochoi (commented):
Agree with @bbroder-algo that we need to change or remove that comment on the client method -- Otherwise LGTM, and I don't have a strong opinion on whether we need to address that falsy problem in this PR.

src/client/v2/indexer/searchForTransactions.ts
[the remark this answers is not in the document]
shiqizng: these remarks are removed.

src/client/v2/indexer/lookupAccountTransactions.ts
```diff
@@ -300,14 +300,12 @@ export default class LookupAccountTransactions extends JSONRequest {
...
    * ```
-   * @remarks
-   * If you are looking for transactions with the currency amount greater than 0, simply construct the query without `currencyGreaterThan` because it doesn't accept `-1`, and passing the `0` `currency-greater-than` value would exclude transactions with a 0 amount.
    *
    * @param greater
    * @category query
    */
   currencyGreaterThan(greater: number) {
-    this.query['currency-greater-than'] = greater;
+    this.query['currency-greater-than'] = greater.toString();
```
algochoi, line 308: nit: I'd add a comment for each of the client methods just in case to make it clear why we're converting to string for now:
```suggestion
    // We convert the following to a string for now to correctly include zero values in request parameters.
    this.query['currency-greater-than'] = greater.toString();
```

algochoi (approved):
Approving - optional nit regarding adding comments for converting params to string

bbroder-algo (approved):

Issue #805, opened 2023-08-03
Add array length check to encodeAddress


winder:
### Subject of the issue

The `encodeAddress` function should check array length before it begins to generate a response. The current implementation would return strange results if the wrong sized array were used. It isn't clear if applications could be depending on this behavior, so we should fix it in the 3.0.0 release.

https://github.com/algorand/js-algorand-sdk/blob/4e1073d6a15b443a73dedf42b025173b397624b0/src/encoding/address.ts#L103

algochoi:
Closed by #809

PR #809, opened 2023-08-08
Add address bytes length check in encodeAddress

algochoi:
Adds check when encoding address to string that address is 32 bytes. Also fixes an example that would fail this check because it passes an empty bytearray to encode. Targets the v3 branch.

https://github.com/algorand/js-algorand-sdk/issues/805

### DIFF
--- a/examples/app.ts
+++ b/examples/app.ts
@@ -153,10 +153,8 @@ async function main() {
 
   // decode b64 string key with Buffer
   const globalKey = algosdk.base64ToString(globalState.key);
-  // decode b64 address value with encodeAddress and Buffer
-  const globalValue = algosdk.encodeAddress(
-    algosdk.base64ToBytes(globalState.value.bytes)
-  );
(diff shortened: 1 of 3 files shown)


examples/app.ts
```diff
@@ -153,10 +153,8 @@ async function main() {
 
   // decode b64 string key with Buffer
   const globalKey = algosdk.base64ToString(globalState.key);
-  // decode b64 address value with encodeAddress and Buffer
-  const globalValue = algosdk.encodeAddress(
-    algosdk.base64ToBytes(globalState.value.bytes)
-  );
+  // show global value
+  const globalValue = globalState.value.bytes;
```
algochoi, line 157: This seemed to be a zero-length byte array in the example test

winder (approved):
LGTM.

Does this SDK have non-cucumber unit tests? If yes, I'd add a test.

PR #811, opened 2023-08-16
CICD: Update chromedriver in package.json to 116+

algochoi:
Updates chromedriver package -- Fixes failing Chrome CI setup.

Extra thoughts:
* I don't think `npm install chromedriver@latest` is necessary, because Chrome 115+ pulls up its own version of chromedriver and use that instead: https://chromedriver.chromium.org/downloads/version-selection ~~Here is an instance where an invalid chromedriver version is supplied (v116 is supplied, but the latest version is [115](https://www.npmjs.com/package/chromedriver)), but the CI process still runs up properly: https://github.com/algorand/js-algorand-sdk/pull/776/commits/7569f079b0ca5ef8021a0c69b2ac31005c391372~~
* I wonder if there's a way to remove this dependency altogether, especially since v115+ installs chromedriver alongside chrome, but this might require custom scripts anyway

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@typescript-eslint/eslint-plugin": "^4.26.1",
     "@typescript-eslint/parser": "^4.26.1",
     "assert": "^2.0.0",
-    "chromedriver": "^114.0.0",
+    "chromedriver": "^115.0.0",
     "concurrently": "^6.2.0",
     "cucumber": "^5.1.0",
     "es-abstract": "^1.18.3",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@typescript-eslint/eslint-plugin": "^4.26.1",
     "@typescript-eslint/parser": "^4.26.1",
     "assert": "^2.0.0",
-    "chromedriver": "^115.0.0",
+    "chromedriver": "^115.0.1",
     "concurrently": "^6.2.0",
     "cucumber": "^5.1.0",
     "es-abstract": "^1.18.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "@typescript-eslint/eslint-plugin": "^4.26.1",
     "@typescript-eslint/parser": "^4.26.1",
     "assert": "^2.0.0",
-    "chromedriver": "^114.0.0",
+    "chromedriver": "^115.0.1",
```
Eric-Warehime, line 41: 116 is latest https://www.npmjs.com/package/chromedriver and that's what I'm seeing when the CI installs the deb.

algochoi: Are we looking at the same page? It say 115.0.1 is the latest tag and it corresponds to what is installed when run: `npm install chromedriver@latest`

It's true that the latest chrome (not the npm package) is 116 on the stable version: https://googlechromelabs.github.io/chrome-for-testing/

algochoi: Double checked: It could be that the changes weren't propagated to npm yet -- there was actually a release on the github repo for 116 after I pushed these changes: https://github.com/giggio/node-chromedriver/tags

I'll be happy to update this after checks pass

algochoi:
@Eric-Warehime My bad, should be updated to 116 now

algobarb (approved):
Approve that it updates chromedriver to 116

Separate note: I wonder if we want to also add installing the latest chromedriver during the create release pr step. Thoughts? https://github.com/algorand/js-algorand-sdk/blob/develop/.github/workflows/create-release-pr.yml#L120

Eric-Warehime (approved):

Issue #821, opened 2023-08-28
atc.gatherSignatures() failing with Multisig


acfunk:
### atc.gatherSignatures() failing with Multisig

When gathering signatures for an AtomicTransactionComposer group with a multisig transaction, the following error is thrown:

`Not enough multisig transactions to merge. Need at least two`

### Your environment

algosdk 2.5.0

### Steps to reproduce

1. Create an ATC with one or more `TransactionWithSigner`s that have a MultiSig signer
2. `const signedTxns = await atc.gatherSignatures()`

### Expected behaviour
`signedTxns` contains signed transactions
### Actual behaviour
error is thrown `Not enough multisig transactions to merge. Need at least two`

winder:
Are you using a multisig that only requires one signature?

acfunk:
Yes, it's a 1 of 2 multi-sig. I started looking into the problem and realized that was a factor. Sorry it wasn't in the repro steps.

I figure something like this in place of this line would fix it...

https://github.com/algorand/js-algorand-sdk/blob/7469c41ab4a54dec44a9ee55da57d54bafd444be/src/signer.ts#L76

      if(partialSigs.length > 1) {
        signed.push(mergeMultisigTransactions(partialSigs));
      } else {
        signed.push(partialSigs[0]);
      }

winder:
@acfunk thanks, that sounds right to me as well. Would you like to make a PR?

acfunk:
@winder I was going to, but I couldn't figure out how to test it locally. Editing the file in node_modules had no effect - too much optimization by Vite it seems. And I'd feel bad submitting a PR that I hadn't tested... unless you told me to.

winder:
@algochoi could you provide some testing guidance here?

algochoi:
We could have a mocha unit test here: https://github.com/algorand/js-algorand-sdk/blob/develop/tests/10.ABI.ts 

something like:
```
it('should accept at least one signature in the multisig', () => {
// ...
});
```

acfunk:
I submitted a PR #822

PR #822, opened 2023-08-29
#821 Only merge multisigs if there are more than 1

acfunk:


algochoi:
I added a unit test here to confirm that this fixes the bug: https://github.com/acfunk/js-algorand-sdk/pull/2

tests/10.ABI.ts
```diff
@@ -1,5 +1,15 @@
 /* eslint-env mocha */
 import assert from 'assert';
+import {
```
algochoi, line 3: Note to any other reviewers: these just organize the imports (they are non-functional changes)

algochoi (approved):
Thanks for the fix!

winder:
Thanks for the contribution @acfunk !

PR #825, opened 2023-09-19
Remove "nonParticipation: true" from offlineKeyReg example

joe-p:
The offlineKeyReg example included `nonParticipation: true` which means the account can no longer participate, which is irreversible. This should not be shown in the example code. 

Also merges in latest develop


jasonpaulos:
This seems like a good fix. However, I don't think this merges in the latest develop. There are more commits in develop after 88aa843.

Particularly, there is ed020a44f3e1b9876dd84cd6f1fbbb208272686b, which updates our node version from 14 to 16. Right now this branch uses 14 and our chromedriver dependency can no longer run (it looks like we may need to update again to node 18 to get it to work).

jasonpaulos:
Fixed by #866

PR #831, opened 2023-09-24
Revert "fetch: Add Cloudflare Workers support (cross-fetch v4) (#794)"

nullun:
This reverts commit 28ba159257c4c7a5732bcf72db9113cd1d03056a as a cheap attempt to fix #823.

PR #833, opened 2023-09-29
fix: remove cross-fetch

### LINKED ISSUE #823: Failed to execute 'fetch' on 'Window': Illegal invocation
### Subject of the issue

Since algosdk v2.5.0 a [sample svelte project](https://github.com/nullun/vite-svelte-algorand) I have now breaks when built and served from a web server, giving the following error message in the console when the page loads.
`Uncaught (in promise) TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation`

A little digging reveals that it is https://github.com/algorand/js-algorand-sdk/pull/794 which causes the issue, and so I believe it's to do with cross-fetch being polyfilled?

The code works if I run `npm run dev` locally in a development environment, it's only broken when I do `npm run build` and serve the dist folder from a web server. The project is currently [deployed](https://vite-svelte-algorand.netlify.app/) using v2.4.0 and it's working as expected, but I'm unable to update to 2.5.0 due to the cross-fetch update.

### Your environment

macOS 13.5.1
node v20.5.1
npm 9.8.0

### Steps to reproduce

1. `git clone git@github.com:nullun/vite-svelte-algorand.git`
2. `cd vite-svelte-algorand`
3. `npm install`
4. `npm run build`
5. Host the `dist` folder on a webserver, I use apache2 locally and have `~/Sites` set to `127.0.0.1/~nullun/`. So I would do something like `ln -s ~/vite-svelte-algorand/dist ~/Sites/dist`
6. Visit http://127.0.0.1/~nullun/dist/ or whatever your destination is

### Expected behaviour

The page loads without errors, showing the latest Algorand Mainnet round.

### Actual behaviour

Page loads with an error, preventing retrieval of latest Algorand Mainnet round.

PhearZero:
# ℹ️ Overview

With the End of Life of Node 16 and engine set to `>=18`, we can safely use native `fetch`.

Consumers targeting `ie11` or other engines can handle the [fetch pollyfill](https://www.npmjs.com/package/whatwg-fetch).  

## 📝 Issues

Fixes #823 
Relates to #831

### DIFF
--- a/tests/cucumber/steps/index.js
+++ b/tests/cucumber/steps/index.js
@@ -54,6 +54,7 @@ if (browser) {
   } else if (browser === 'chrome') {
     require('chromedriver');
     browserHeaders = [
+      'accept-language',
       'origin',
       'referer',
       'sec-ch-ua',
@@ -77,6 +78,7 @@ if (browser) {
 
   console.log('Testing in browser');
 } else {
+  browserHeaders = ['sec-fetch-mode', 'accept-language'];
   console.log('Testing in node');
 }
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,6 @@
 module.exports = {
   env: {
+    browser: true,
     es2021: true,
     'shared-node-browser': true,
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
       "dependencies": {
         "algo-msgpack-with-bigint": "^2.1.1",
         "buffer": "^6.0.3",
-        "cross-fetch": "^4.0.0",
         "hi-base32": "^0.5.1",
         "js-sha256": "^0.9.0",
         "js-sha3": "^0.8.0",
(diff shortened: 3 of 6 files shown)


tests/cucumber/steps/index.js
```diff
@@ -54,6 +54,7 @@ if (browser) {
   } else if (browser === 'chrome') {
     require('chromedriver');
     browserHeaders = [
+      'accept-language',
```
jasonpaulos, line 57: In CI this test is failing for chrome because `accept-language` isn't present. Did this behavior differently for you locally?

PhearZero: `TEST_BROWSER=chrome make unit` is passing locally for Ubuntu 20.04.3 LTS using `v18.17.1` and `v20.5.0`

jasonpaulos: Got it. I believe this test is too brittle for reliable use. Could you remove the `@unit.client-no-headers` line from `unit.tags` in this PR?

I will create a separate issue to fix the test (it likely should be checking that certain headers aren't present, not that all present headers must match a hardcoded list).

nullun:
Just want to confirm that this PR fixes #823 and I'll close my revert (#831).

PhearZero:
> Just want to confirm that this PR fixes #823 and I'll close my revert (#831).

@nullun Just deployed an example with the following dependency in `package.json`:

```json
{
  "algosdk": "github:PhearZero/js-algorand-sdk#fix/cross-fetch"
}
```

And the current render of `getTransactionParams` : https://wrangler-test.phearzero.workers.dev/

jasonpaulos (approved):
Great work, happy to remove this dependency

PR #834, opened 2023-10-05
TEAL Source Map: Improve SourceMap and support columns

jasonpaulos:
Improve the source map class and take advantage of column reporting from https://github.com/algorand/go-algorand/pull/5776 to provide more accurate mappings.

Also, rename from `SourceMap` to `ProgramSourceMap` to be more specific. Our source maps are not exactly the same as general source maps, since they map indexes from a binary file (the compiled bytecode) to source locations.

### DIFF
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="sourcemap-v2-testcases"
+SDK_TESTING_BRANCH="V2"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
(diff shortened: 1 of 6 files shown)


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="V2"
+SDK_TESTING_BRANCH="sourcemap-v2-testcases"
```
jasonpaulos, line 3: TODO: revert back

jannotti (approved):
Seems good to me. I don't follow the release process as closely as I should - is this name change "SourceMap" to "ProgramSourceMap" happening across the v2/v3 boundary?

And, if this is indeed going into v3, maybe call the new sourcemap tests "sourcemapv3" even though it skips a version number, to be in sync?  Though I suppose that is not user facing.

jasonpaulos:
Correct this would be released in v3.

I don't love the `sourcemapsv2` name, but we recently moved from v1 to v2 of the cucumber testing features, which is what I took inspiration from.

.test-env
jasonpaulos:
```suggestion
SDK_TESTING_BRANCH="V2"
```

PR #820, opened 2023-08-28
chore: convert to esm

PhearZero:
# ℹ️ Overview

This is an experimental PR that sees how difficult it would be to convert to native `esm`.

- Reproduction of Issue: https://github.com/PhearZero/js-algorand-sdk-node-18-module/tree/main
- Working Fix with ESM: https://github.com/PhearZero/js-algorand-sdk-node-18-module/tree/fix/esm
- Working Fix with CJS: https://github.com/PhearZero/js-algorand-sdk-node-18-module/tree/fix/cjs

### Remaining Items:

- [x] mocha
- [ ] cucumber

### 📝 Issues
- #819

PR #836, opened 2023-10-16
chore: native esm bundle

### LINKED ISSUE #819: Native ESM
##  ❌ Problem

The following errors due to `node.js` resolving the `commonjs` bundle while using native `esm` modules. 

Reproduction: https://github.com/PhearZero/js-algorand-sdk-node-18-module/tree/main

```javascript
import { Indexer } from 'algosdk'
```

### 🤔Reasoning:

While modern bundlers have robust resolvers, it would be good to have first class support for `esm` without third party tools. This will also improve third party bundlers ability to tree shake. 

## ✅ Solution

Convert package to `esm`.

- Package type "commonjs" loads the `./dist/cjs/index.js` entrypoint. ([fixed example using fork](https://github.com/PhearZero/js-algorand-sdk-node-18-module/tree/fix/cjs))
- Package type "module" loads the `./dist/esm/index.js` entrypoint. ([fixed example using fork](https://github.com/PhearZero/js-algorand-sdk-node-18-module/tree/fix/esm))

## 📦Dependencies

This requires moving to `tsx` and slight refactoring of the codebase. 

##  ‼️ Urgency

Very low priority

PhearZero:
# Overview

- Allows for native esm imports of `algosdk`
- Configures exports field of `package.json`
- Uses `TSX` as a replacement for `ts-node`

# Remaining
- [x] Validate third party type consumers
- [x] Validate happy path type inference 
- [x] Refine exports field

# Issues
- fixes #819 
- related to #820


package.json
```diff
@@ -8,6 +8,98 @@
     ".": "dist/browser/algosdk.min.js",
     "crypto": false
   },
+  "exports": {
+    ".": {
+      "import": "./dist/esm/index.js",
+      "require": "./dist/cjs/index.js",
+      "types": "./dist/types/index.d.ts"
+    },
```
jasonpaulos, line 16: I'm curious why you added more export paths beyond just this one. Do you think they are necessary/useful?

Our mindset has been to expose everything public through the top-level import, since that allows us to make changes to the package internal file structure without breaking anyone's imports.

I guess I'm just saying I'd rather not encourage/allow additional import paths, unless there is a compelling reason to do so.

PhearZero: Completely agree, it shouldn't be encouraged unless there is a good reason. I've added them all just for completeness and we remove everything other than the critical ones. The main reasons are tree-shaking and when type inference fails. 

## Critical exports

The most important one is the `HTTPClient` models, it's very likely that we lose the inference of the original type and we need to `cast` the response type in third party libraries. 

## Nice to haves

- `client` Perfect target for tree-shaking. In projects without any interactivity they either use the `HTTPClients` or write their own custom `fetch` interfaces. Having this be stand-alone would reduce third party bundle sizes in this context. A lot of projects allow for the barrel file export (single entrypoint) along side the tree-shakeable export path

- `nacl/encoding` would be a good option for third parties to implement their own transaction factories using the custom encoding and nacl wrapper. If the third party implementation are sound, it would be much easier to adopt in the `sdk`

jasonpaulos (approved):
I fixed the error in the cucumber browser tests and made some minor tweaks to a few exports, and this looks mergable now. Thanks for putting in the effort to make this PR

PR #843, opened 2023-12-01
feature: tsup builds

RobbyUitbeijerse:
I've been using [tsup](https://github.com/egoist/tsup) for a while to create esbuild based builds a lot faster than Webpack ever could. The webpack configuration is pretty simple, allowing for a smooth transition.

Noteworthy: the dist output structure does change with this switch.

There are a couple of things that need to happen before this is ready. Let me know if you're interested so I can take a look.

1. The mocha test runner is depending on the webpack set-up, I think this could easily be moved to the [web test runner](https://modern-web.dev/docs/test-runner/overview/)
2. Documentation changes about the changing dist directory structure
3. Adjust the ci/cd tests 
4. Will need to properly the tweetnacl-js case that is handled in the current webpack set-up


emg110:
Tsup is a great tool but Are you proposing the addition of support for tsup or is it replacing webpack with tsup? If that's the latter then may I humbly ask why do you think this big change should happen (I mean which critical issue or shortcoming does this change address or solve?) and what would be the benefits? 
Because there are many different build tools out there and if every developer PRs their own prefered method into an opensource library , well that would not be ideal (all IMHO)!
Plus , Tsup has some road to walk before becoming 100% ready (style loaders are still incomplete and experimental) for all-purpose production and does not enjoy the vast plugin ecosystem that already exists around projects like Webpack and Rollup!

RobbyUitbeijerse:
Hey! That's fair - I'm not aiming to solve any existing problem but rather try to share tools and experience that I came across throughout time. This year I've moved most of my preferred toolchain to go/rust based tools, another example of that being the [biome PR](https://github.com/algorand/js-algorand-sdk/pull/842). These tools don't solve an existing issue on their own as much as they allow me and my team to enjoy programming more on a day to day basis

Throughout time, the development community moved from Bower to npm and Gulp to Webpack. Those transitions wouldn't have happened without people making an effort of improving the things they're building. The same goes for [one of the existing pull requests](https://github.com/algorand/js-algorand-sdk/pull/836), which moves `ts-node` to `tsx` , and for people moving from npm to pnpm or yarn.

Regarding tsup specifically; it have been solid me and due to the use of esbuild is **remarkably** faster than webpack while only requiring a single dependency to both compile whatever you want and generate types. The style loading comment is fair - however as this is a js sdk, so while it's true - it's not quite relevant here. Let me know if you're interested, also happy to move it to e.g. SWC or close the PR.

PR #852, opened 2024-01-10
3.0.0: Remove `IntDecoding` as a REST option & support native bigint types in models

### LINKED ISSUE #848: Remove int decoding option, support bigint by default
Remove the `IntDecoding` enum option on REST API calls. This option was originally introduced to allow bigints to be decoded by the REST API in a backwards-compatible way, but it's no longer needed in v3.

Instead, we should decode with support for bigints by default (either by upgrading every int to a bigint or only those which exceed `Number.MAX_SAFE_INTEGER`). The algod and indexer REST API specs _should_ indicate which fields actually need to be represented by a uint64 or larger with the `"x-algorand-format": "uint64"` value; these can be typed as bigint by this SDK, and the rest can remain numbers.

jasonpaulos:
Relies on https://github.com/algorand/generator/pull/72

Fixes #848 (that issue contains a good description of this work)

### DIFF
--- a/examples/accounts.ts
+++ b/examples/accounts.ts
@@ -101,7 +101,7 @@ async function main() {
   const acctInfo = await client.accountInformation(acct1.addr).do();
 
   console.log(
-    `Account Info: ${JSON.stringify(acctInfo)} Auth Addr: ${
+    `Account Info: ${algosdk.stringifyJSON(acctInfo)} Auth Addr: ${
       acctInfo['auth-addr']
     }`
   );
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -224,24 +224,32 @@ export class HTTPClient {
 
   /**
    * Send a GET request.
-   * @param relativePath - The path of the request.
-   * @param jsonOptions - Options object to use to decode JSON responses. See
+   *
+   * @param options - The options to use for the request.
+   * @param options.relativePath - The path of the request.
+   * @param options.jsonOptions - Options object to use to decode JSON responses. See
    *   utils.parseJSON for the options available.
-   * @param query - An object containing the query parameters of the request.
-   * @param requestHeaders - An object containing additional request headers to use.
-   * @param parseBody - An optional boolean indicating whether the response body should be parsed
+   * @param options.query - An object containing the query parameters of the request.
+   * @param options.requestHeaders - An object containing additional request headers to use.
+   * @param options.parseBody - An optional boolean indicating whether the response body should be parsed
    *   or not.
    * @returns Response object.
    */
-  async get(
-    relativePath: string,
-    jsonOptions: utils.ParseJSONOptions,
-    query?: Query<any>,
-    requestHeaders: Record<string, string> = {},
-    parseBody: boolean = true
-  ): Promise<HTTPClientResponse> {
+  async get({
+    relativePath,
+    jsonOptions,
+    query,
+    requestHeaders,
+    parseBody,
+  }: {
+    relativePath: string;
+    jsonOptions: utils.ParseJSONOptions;
+    query?: Query<any>;
+    requestHeaders?: Record<string, string>;
+    parseBody: boolean;
+  }): Promise<HTTPClientResponse> {
     const format = getAcceptFormat(query);
-    const fullHeaders = { ...requestHeaders, accept: format };
+    const fullHeaders = { ...(requestHeaders ?? {}), accept: format };
 
     try {
       const res = await this.bc.get(
@@ -260,18 +268,26 @@ export class HTTPClient {
    * Send a POST request.
    * If no content-type present, adds the header "content-type: application/json"
    * and data is serialized in JSON (if not empty)
+   * @param options - The options to use for the request.
    */
-  async post(
-    relativePath: string,
-    data: any,
-    jsonOptions: utils.ParseJSONOptions,
-    query?: Query<any>,
-    requestHeaders: Record<string, string> = {},
-    parseBody: boolean = true
-  ): Promise<HTTPClientResponse> {
+  async post({
+    relativePath,
+    data,
+    jsonOptions,
+    query,
+    requestHeaders,
+    parseBody,
+  }: {
+    relativePath: string;
+    data: any;
+    jsonOptions: utils.ParseJSONOptions;
+    query?: Query<any>;
+    requestHeaders?: Record<string, string>;
+    parseBody: boolean;
+  }): Promise<HTTPClientResponse> {
     const fullHeaders = {
       'content-type': 'application/json',
-      ...tolowerCaseKeys(requestHeaders),
+      ...tolowerCaseKeys(requestHeaders ?? {}),
     };
 
     try {
@@ -297,17 +313,24 @@ export class HTTPClient {
    * Send a DELETE request.
    * If no content-type present, adds the header "content-type: application/json"
    * and data is serialized in JSON (if not empty)
+   * @param options - The options to use for the request.
    */
-  async delete(
-    relativePath: string,
-    data: any,
-    jsonOptions: utils.ParseJSONOptions,
-    requestHeaders: Record<string, string> = {},
-    parseBody: boolean = true
-  ) {
+  async delete({
+    relativePath,
+    data,
+    jsonOptions,
+    requestHeaders,
+    parseBody,
+  }: {
+    relativePath: string;
+    data: any;
+    jsonOptions: utils.ParseJSONOptions;
+    requestHeaders?: Record<string, string>;
+    parseBody: boolean;
+  }) {
     const fullHeaders = {
       'content-type': 'application/json',
-      ...tolowerCaseKeys(requestHeaders),
+      ...tolowerCaseKeys(requestHeaders ?? {}),
     };
 
     try {
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -112,8 +112,7 @@ export class HTTPClient {
   }
 
   /**
-   * Parse JSON using either the built-in JSON.parse or utils.parseJSON
-   * depending on whether jsonOptions are provided or not
+   * Parse JSON using utils.parseJSON
    *
    * @param text - JSON data
    * @param status - Status of the response (used in case parseJSON fails)
@@ -123,13 +122,13 @@ export class HTTPClient {
   public static parseJSON(
     text: string,
     status: number,
-    jsonOptions: utils.JSONOptions = {}
+    jsonOptions: utils.ParseJSONOptions
   ) {
     try {
-      if (Object.keys(jsonOptions).length === 0) {
-        return text && JSON.parse(text);
+      if (!text) {
+        return null;
       }
-      return text && utils.parseJSON(text, jsonOptions);
+      return utils.parseJSON(text, jsonOptions);
     } catch (err_) {
       const err = err_ as ErrorWithAdditionalInfo;
       // return the raw response if the response parsing fails
@@ -156,7 +155,7 @@ export class HTTPClient {
       return new Uint8Array(0); // empty Uint8Array
     }
     if (requestHeaders['content-type'] === 'application/json') {
-      return new TextEncoder().encode(JSON.stringify(data));
+      return new TextEncoder().encode(utils.stringifyJSON(data));
     }
     if (typeof data === 'string') {
       return new TextEncoder().encode(data);
@@ -178,7 +177,7 @@ export class HTTPClient {
     res: BaseHTTPClientResponse,
     format: 'application/msgpack' | 'application/json',
     parseBody: boolean,
-    jsonOptions: utils.JSONOptions = {}
+    jsonOptions: utils.ParseJSONOptions
   ): HTTPClientResponse {
     let { body } = res;
     let text: string | undefined;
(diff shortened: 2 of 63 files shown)


src/client/kmd.ts
```diff
@@ -89,7 +117,14 @@ export class KmdClient extends ServiceClient {
...
     };
-    const res = await this.c.post('/v1/wallet/release', req);
+    const res = await this.c.post({
+      relativePath: '/v1/wallet/release',
+      data: req,
+      parseBody: true,
+      jsonOptions: {
+        intDecoding: IntDecoding.SAFE,
+      },
+    });
```
jannotti, line 127: Perhaps `KmdClient` should have a `post` method with `parseBody: true` and the `jsonOptions` set?  Maybe even just take two arguments, `relativePath` and `data`, so all of these stay as trivial as they were before by calling `this.post` instead of `this.c.post`?
I may delete this if/when I notice there are exceptions.

src/dryrun.ts
```diff
@@ -148,79 +149,7 @@ export async function createDryrun({
   });
 }
 
-interface StackValueResponse {
```
gmalouf, line 151: This file's changes feel somewhat unrelated to the stated goal of the PR, can you walk me through the intentions here?

src/client/v2/algod/models/types.ts
```diff
@@ -624,9 +637,9 @@ export class AccountParticipation extends BaseModel {
       typeof selectionParticipationKey === 'string'
         ? base64ToBytes(selectionParticipationKey)
         : selectionParticipationKey;
-    this.voteFirstValid = voteFirstValid;
-    this.voteKeyDilution = voteKeyDilution;
-    this.voteLastValid = voteLastValid;
+    this.voteFirstValid = ensureBigInt(voteFirstValid);
+    this.voteKeyDilution = ensureBigInt(voteKeyDilution);
+    this.voteLastValid = ensureBigInt(voteLastValid);
```
jannotti, line 642: Just so I understand how much scrutiny is required - all these changes were forced upon you by the compiler, right?  You didn't have to catch each required use of `ensureBigInt` by hand, correct?

src/client/v2/algod/models/types.ts
```diff
@@ -1106,7 +1119,10 @@ export class ApplicationParams extends BaseModel {
         ? base64ToBytes(clearStateProgram)
         : clearStateProgram;
     this.creator = creator;
-    this.extraProgramPages = extraProgramPages;
+    this.extraProgramPages =
+      typeof extraProgramPages === 'undefined'
+        ? undefined
+        : ensureSafeInteger(extraProgramPages);
```
jannotti, line 1125: Why wouldn't we want to coerce to 0 instead of `undefined`?  The question probably comes from me not having enough context to know where this type is used.

src/client/v2/algod/models/types.ts
```diff
@@ -1301,15 +1317,15 @@ export class ApplicationStateSchema extends BaseModel {
...
   ): ApplicationStateSchema {
     /* eslint-disable dot-notation */
-    if (typeof data['num-uint'] === 'undefined')
-      throw new Error(`Response is missing required field 'num-uint': ${data}`);
     if (typeof data['num-byte-slice'] === 'undefined')
       throw new Error(
         `Response is missing required field 'num-byte-slice': ${data}`
       );
+    if (typeof data['num-uint'] === 'undefined')
+      throw new Error(`Response is missing required field 'num-uint': ${data}`);
```
jannotti, line 1325: Why are these required, rather than defaulting to 0?

src/client/v2/algod/models/types.ts
```diff
@@ -2738,8 +2754,14 @@ export class DryrunTxnResult extends BaseModel {
     this.disassembly = disassembly;
     this.appCallMessages = appCallMessages;
     this.appCallTrace = appCallTrace;
-    this.budgetAdded = budgetAdded;
-    this.budgetConsumed = budgetConsumed;
+    this.budgetAdded =
+      typeof budgetAdded === 'undefined'
+        ? undefined
+        : ensureSafeInteger(budgetAdded);
```
jannotti, line 2760: This is a repeat of the previous question (why not coerce to 0). But more broadly, could `ensureSafeInteger` accept `undefined` and return 0?

src/client/v2/algod/models/types.ts
```diff
@@ -4571,7 +4678,7 @@ export class SimulateUnnamedResourcesAccessed extends BaseModel {
    * The unnamed applications that were referenced. The order of this array is
    * arbitrary.
    */
-  public apps?: (number | bigint)[];
```
jannotti, line 4574: Oh, this is allowed?  I like that better than the Array stuff when an element is a union.

src/client/v2/algod/models/types.ts
```diff
@@ -4631,11 +4738,16 @@ export class SimulateUnnamedResourcesAccessed extends BaseModel {
     super();
     this.accounts = accounts;
     this.appLocals = appLocals;
-    this.apps = apps;
+    this.apps =
+      typeof apps === 'undefined' ? undefined : apps.map(ensureBigInt);
```
jannotti, line 4742: I'm repeating myself somewhat as I think I'm understanding the issue better.  I think we ought to coercing all of the "omitEmpty" fields, to their "empty" values. So lists become `[]`, integers become `0` or `0n`.  I've only been commenting on the places you've had to edit because of the `bignum` change, but this seems like something we ought to do for all of these types before handing them to the user. _They_ shouldn't have to understand omitEmpty.

src/client/v2/algod/setSyncRound.ts
```diff
@@ -1,22 +1,27 @@
...
+    super(c);
   }
 
   path() {
     return `/v2/ledger/sync/${this.round}`;
   }
 
   async do(headers = {}) {
-    const res = await this.c.post(this.path(), null, undefined, headers);
+    const res = await this.c.post({
```
jannotti, line 18: Similar to the `KmdClient.post` suggestion but deeper down, should we change `HTTPClient` to make these call sites simpler, since they are the common case.  Seems like `parseBody` and `jsonOptions` should default to these, so you can go back to simple call sites for most api calls (or add a new method like `apiPost`)

src/types/intDecoding.ts
```diff
@@ -6,7 +6,7 @@ enum IntDecoding {
    * All integers will be decoded as Numbers, meaning any values greater than
    * Number.MAX_SAFE_INTEGER will lose precision.
    */
-  DEFAULT = 'default',
+  UNSAFE = 'unsafe',
```
jannotti, line 9: github won't let me comment in the right place.  Are we still using `MIXED`? I get the impression that you don't need it because you can use `bigint`, then your coerce routines will fix things up.

jannotti (commented):
I don't think I understand how everything fits together, but here are some comments before I'm ooo.

src/client/v2/algod/models/types.ts
gmalouf: is this syntax for an empty array if not specified?

gmalouf (commented):
I'd like to see discussion to JJ's questions around the right default types for not set/not existing (nils vs zero values). Outside of that, this largely aligns with your description of the issue.

src/client/v2/algod/models/types.ts
jasonpaulos, line 1: I meant to comment this earlier, but this file is generated by https://github.com/algorand/generator. See https://github.com/algorand/generator/pull/72 for the changes to the template

src/client/v2/indexer/models/types.ts
jasonpaulos, line 1: This is also generated in the same way: https://github.com/algorand/generator/pull/72

src/types/intDecoding.ts
jasonpaulos: I don't think we use currently using MIXED anywhere, but it's still a potentially useful option to have, especially since we now expose the JSON parsing to users

jannotti: I don't think I followed why we should expose the json parsing to users. Is the idea that they may not want our models, but want to use the sdk to hit an endpoint then parse it (differently) themselves?

examples/accounts.ts
```diff
@@ -101,7 +101,7 @@ async function main() {
   const acctInfo = await client.accountInformation(acct1.addr).do();
 
   console.log(
-    `Account Info: ${JSON.stringify(acctInfo)} Auth Addr: ${
+    `Account Info: ${algosdk.stringifyJSON(acctInfo)} Auth Addr: ${
```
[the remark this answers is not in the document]
jasonpaulos: My experience is that, yes, `console.log("Account Info", acctInfo)` would expand `acctInfo` for you in browsers, and it's usually a drop down that you can interact with to see each field. However, if you're not on a browser, it's not a great experience as often you'll just get back a string like `[object, object]`.

src/client/kmd.ts
jasonpaulos: Not a bad idea, all of the POST requests here are very similar

src/client/v2/algod/models/types.ts
jasonpaulos: My bad, I meant to add https://github.com/algorand/js-algorand-sdk/pull/852#discussion_r1491320035 before requesting reviews -- this file is generated, not handwritten

src/client/v2/algod/models/types.ts
jasonpaulos: The REST API declares them as required

src/client/v2/algod/models/types.ts
jasonpaulos: `ensureSafeInteger(undefined)` will error. If you wanted it to coerce undefined to 0, you could do so with `ensureSafeInteger(x ?? 0)` (the `??` being [the null coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing), which is relatively new). The `Transaction` constructor uses this pattern a lot

src/client/v2/algod/models/types.ts
jasonpaulos: It is allowed and equivalent to `Array<number | bigint>`. I thought our formatter preferred the `Array<*>` syntax, but if this was here I'm doubting that now

src/client/v2/algod/models/types.ts
jasonpaulos: I agree with this a lot. I'll think more about how this can be addresses

src/client/v2/algod/setSyncRound.ts
jasonpaulos: I'm also unhappy with the `HTTPClient` methods at the moment, though I didn't know whether I should address them in this PR.

My inclination is to simplify those methods so that they just return the raw `Uint8Array` bytes, and perhaps introduce new methods which would return parsed JSON.

So it would be something like `HTTPClient.get` and `HTTPClient.getJSON` -- still thinking through this

src/client/v2/algod/models/types.ts
jannotti: Required in the _response_? (I'm judging based on the error string.)

Seems odd that the SDK doesn't trust server side output in this one particular way.

src/client/v2/algod/models/types.ts
jannotti: Well, I'll resolve. After all, it's deleted code.

src/client/v2/algod/models/types.ts
jannotti: It feels like it can be another PR, but would be very nice to do for v3.

src/dryrun.ts
jasonpaulos: Good question. This code broke because algod responses now contain bigints for certain fields. I could have just changed those fields, but I noticed a lot of the definitions here were redundant and no longer necessary, since we have fully typed algod responses.

src/client/kmd.ts
jasonpaulos: Added in 80a0c35

src/types/intDecoding.ts
jasonpaulos: Right now it's expected that many of our type representations can be encoded into JSON and msgpack and back, and not just by our code -- we expect the user to be able to perform these encoding operations if they want to persist things to disk, for instance. Many of our types have a `get_object_for_encoding` method which returns an object ready to be serialized into one of those formats. Unfortunately bigints aren't supported by the standard JSON encoder, so without exposing the additional utilities, there wouldn't be an easy way to encode these newer objects in v3.

Technically this was also a problem in v2, but due to how types stored internal representations (they'd store exactly what they got most of the time), it was only a problem if you created a type with a bigint.

src/client/client.ts
```diff
@@ -221,24 +224,32 @@ export class HTTPClient {
 
   /**
    * Send a GET request.
-   * @param relativePath - The path of the request.
```
algorandskiy, line 224: isn't it changing v2 client as well? Or /v2/client calls these methods internally? Not sue how it is structured.

src/client/v2/indexer/models/types.ts
```diff
@@ -3,6 +3,7 @@
  */
 
 /* eslint-disable no-use-before-define */
+import { ensureBigInt, ensureSafeInteger } from '../../../../utils/utils.js';
```
algorandskiy, line 6: do we really want to re-generate v2 client models?

algorandskiy (commented):
LGTM but not sure about v2 client backward compatibility, please advise.

src/client/client.ts
jasonpaulos: Every HTTP request goes through these functions. To be clear, we are still supporting the v2 algod and indexer APIs -- JS v3 is upgrading the library version, not the REST API versions

src/client/v2/indexer/models/types.ts
jasonpaulos: Yes this brings some improvements. Before every integer field was `number | bigint`, which is not useful because you'd need to cast to one of them before actually comparing or operating on them

src/client/client.ts
gmalouf: This is a purely a js client upgrade, APIs should be stable.

gmalouf (approved):
I'm good at this point.

algorandskiy (approved):

PR #856, opened 2024-02-17
feat: add ARC22 and ARC28 interfaces for ABI contracts and methods

joe-p:
Adds ARC28 (events) and ARC22 (readonly) to ABI-related interfaces

### DIFF
--- a/src/abi/interface.ts
+++ b/src/abi/interface.ts
@@ -1,4 +1,5 @@
 import { ABIMethod, ABIMethodParams, getMethodByName } from './method';
+import { ARC28Event } from './event';
 
 export interface ABIInterfaceParams {
   name: string;
@@ -10,6 +11,8 @@ export class ABIInterface {
   public readonly name: string;
   public readonly description?: string;
   public readonly methods: ABIMethod[];
+  /** [ARC-28](https://arc.algorand.foundation/ARCs/arc-0028) events that MAY be emitted by this contract */
+  public readonly events?: ARC28Event[];
 
   constructor(params: ABIInterfaceParams) {
     if (typeof params.name !== 'string' || !Array.isArray(params.methods)) {
--- a/src/abi/interface.ts
+++ b/src/abi/interface.ts
@@ -1,5 +1,4 @@
 import { ABIMethod, ABIMethodParams, getMethodByName } from './method';
-import { ARC28Event } from './event';
 
 export interface ABIInterfaceParams {
   name: string;
@@ -11,8 +10,6 @@ export class ABIInterface {
   public readonly name: string;
   public readonly description?: string;
   public readonly methods: ABIMethod[];
-  /** [ARC-28](https://arc.algorand.foundation/ARCs/arc-0028) events that MAY be emitted by this contract */
-  public readonly events?: ARC28Event[];
 
   constructor(params: ABIInterfaceParams) {
     if (typeof params.name !== 'string' || !Array.isArray(params.methods)) {
--- a/src/abi/method.ts
+++ b/src/abi/method.ts
@@ -2,6 +2,7 @@ import { genericHash } from '../nacl/naclWrappers';
 import { ABIType, ABITupleType } from './abi_type';
 import { ABITransactionType, abiTypeIsTransaction } from './transaction';
 import { ABIReferenceType, abiTypeIsReference } from './reference';
+import { ARC28Event } from './event';
 
 function parseMethodSignature(
   signature: string
@@ -61,6 +62,10 @@ export interface ABIMethodParams {
   desc?: string;
   args: ABIMethodArgParams[];
   returns: ABIMethodReturnParams;
+  /** Optional, is it a read-only method (according to [ARC-22](https://arc.algorand.foundation/ARCs/arc-0022)) */
+  readonly?: boolean;
+  /** [ARC-28](https://arc.algorand.foundation/ARCs/arc-0028) events that MAY be emitted by this method */
+  events?: ARC28Event[];
 }
 
 export type ABIArgumentType = ABIType | ABITransactionType | ABIReferenceType;
--- a/src/abi/method.ts
+++ b/src/abi/method.ts
@@ -82,7 +82,8 @@ export class ABIMethod {
   }>;
 
   public readonly returns: { type: ABIReturnType; description?: string };
-  public readonly events: ARC28Event[];
+  public readonly events?: ARC28Event[];
+  public readonly readonly?: boolean;
 
   constructor(params: ABIMethodParams) {
     if (
@@ -118,7 +119,8 @@ export class ABIMethod {
       description: params.returns.desc,
     };
 
-    this.events = params.events || [];
+    this.events = params.events;
+    this.readonly = params.readonly;
   }
 
   getSignature(): string {
@@ -156,6 +158,7 @@ export class ABIMethod {
         desc: this.returns.description,
       },
       events: this.events,
+      readonly: this.readonly,
     };
   }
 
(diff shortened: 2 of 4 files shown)


joe-p:
I'm not sure what happened to this PR... I may need to make a new one

Running a diff locally against upstream `develop` shows

```
diff --git a/src/abi/contract.ts b/src/abi/contract.ts
index d4d6236..3e0ed81 100644
--- a/src/abi/contract.ts
+++ b/src/abi/contract.ts
@@ -1,4 +1,5 @@
 import { ABIMethod, ABIMethodParams, getMethodByName } from './method';
+import { ARC28Event } from './event';
 
 export interface ABIContractNetworkInfo {
   appID: number;
@@ -20,6 +21,8 @@ export class ABIContract {
   public readonly description?: string;
   public readonly networks: ABIContractNetworks;
   public readonly methods: ABIMethod[];
+  /** [ARC-28](https://arc.algorand.foundation/ARCs/arc-0028) events that MAY be emitted by this contract */
+  public readonly events?: ARC28Event[];
 
   constructor(params: ABIContractParams) {
     if (
diff --git a/src/abi/event.ts b/src/abi/event.ts
new file mode 100644
index 0000000..ecdcb84
--- /dev/null
+++ b/src/abi/event.ts
@@ -0,0 +1,15 @@
+export interface ARC28Event {
+  /** The name of the event */
+  name: string;
+  /** Optional, user-friendly description for the event */
+  desc?: string;
+  /** The arguments of the event, in order */
+  args: Array<{
+    /** The type of the argument */
+    type: string;
+    /** Optional, user-friendly name for the argument */
+    name?: string;
+    /** Optional, user-friendly description for the argument */
+    desc?: string;
+  }>;
+}
diff --git a/src/abi/interface.ts b/src/abi/interface.ts
index b676c5d..b9f67a7 100644
--- a/src/abi/interface.ts
+++ b/src/abi/interface.ts
@@ -1,4 +1,5 @@
 import { ABIMethod, ABIMethodParams, getMethodByName } from './method';
+import { ARC28Event } from './event';
 
 export interface ABIInterfaceParams {
   name: string;
@@ -10,6 +11,8 @@ export class ABIInterface {
   public readonly name: string;
   public readonly description?: string;
   public readonly methods: ABIMethod[];
+  /** [ARC-28](https://arc.algorand.foundation/ARCs/arc-0028) events that MAY be emitted by this contract */
+  public readonly events?: ARC28Event[];
 
   constructor(params: ABIInterfaceParams) {
     if (typeof params.name !== 'string' || !Array.isArray(params.methods)) {
diff --git a/src/abi/method.ts b/src/abi/method.ts
index 00704ef..d939a1f 100644
--- a/src/abi/method.ts
+++ b/src/abi/method.ts
@@ -2,6 +2,7 @@ import { genericHash } from '../nacl/naclWrappers';
 import { ABIType, ABITupleType } from './abi_type';
 import { ABITransactionType, abiTypeIsTransaction } from './transaction';
 import { ABIReferenceType, abiTypeIsReference } from './reference';
+import { ARC28Event } from './event';
 
 function parseMethodSignature(
   signature: string
@@ -61,6 +62,10 @@ export interface ABIMethodParams {
   desc?: string;
   args: ABIMethodArgParams[];
   returns: ABIMethodReturnParams;
+  /** Optional, is it a read-only method (according to [ARC-22](https://arc.algorand.foundation/ARCs/arc-0022)) */
+  readonly?: boolean;
+  /** [ARC-28](https://arc.algorand.foundation/ARCs/arc-0028) events that MAY be emitted by this method */
+  events?: ARC28Event[];
 }
 
 export type ABIArgumentType = ABIType | ABITransactionType | ABIReferenceType;

```

pbennett:
Just chiming in to say this would be nice to get in - it's preventing me from cleanly including ARC28 events in my tealscript contracts.  The generated json blocks tests from being run because it doesn't match the schema defined in the sdk.

src/abi/interface.ts
```diff
@@ -10,6 +11,8 @@ export class ABIInterface {
   public readonly name: string;
   public readonly description?: string;
   public readonly methods: ABIMethod[];
+  /** [ARC-28](https://arc.algorand.foundation/ARCs/arc-0028) events that MAY be emitted by this contract */
+  public readonly events?: ARC28Event[];
```
jasonpaulos, line 15: I don't see anything in the ARC-28 spec about events being associated with interfaces. I only them as a property of either Contracts or Methods, so this feels like it shouldn't be here

src/abi/method.ts
```diff
@@ -77,6 +82,7 @@ export class ABIMethod {
   }>;
 
   public readonly returns: { type: ABIReturnType; description?: string };
+  public readonly events: ARC28Event[];
```
jasonpaulos, line 85: You're still missing readonly

jasonpaulos:
Also it looks like this failing in cucumber tests:
```
13) Scenario: Serialize Contract into json # tests/cucumber/features/unit/abijson.feature:97
    ✔ When I create the Method object from method signature "add(uint32,uint32)uint32" # tests/cucumber/steps/index.js:510
    ✔ And I create a Contract object from the Method object with name "ExampleContract" and description "This is an example contract" # tests/cucumber/steps/index.js:510
    ✔ And I set the Contract's appID to 1234 for the network "wGHE2Pwdvd7S12BL5FaOP20EGYesN73ktiC1qzkkit8=" # tests/cucumber/steps/index.js:510
    ✔ And I set the Contract's appID to 5678 for the network "SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=" # tests/cucumber/steps/index.js:510
    ✔ And I serialize the Contract object into json # tests/cucumber/steps/index.js:510
    ✖ Then the produced json should equal "contract.json" loaded from "abi_responsejsons" # tests/cucumber/steps/index.js:579
        AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
        + actual - expected ... Lines skipped

          {
            desc: 'This is an example contract',
        +   events: [],
        +   methods: [
        +     {
        +       args: [
        +         {
        +           type: 'uint32'
        +         },
        +         {
        +           type: 'uint32'
        +         }
        +       ],
        +       events: [],
        -   methods: [
        -     {
        -       args: [
        -         {
        -           type: 'uint32'
        -         },
        -         {
        -           type: 'uint32'
        -         }
        -       ],
                name: 'add',
        ...
              }
            }
          }
            + expected - actual

             {
               "desc": "This is an example contract"
            -  "events": []
               "methods": [
                 {
                   "args": [
                     {
                     {
                       "type": "uint32"
                     }
                   ]
            -      "events": []
                   "name": "add"
                   "returns": {
                     "type": "uint32"
                   }
```
Perhaps when converting to JSON, if no events are present, that field should be omitted from the result? Since these tests run on every SDK, they're a bit difficult to deal with.

joe-p:
I made `readonly` and `events` optional so if they weren't provided upon instantiation they won't be included in the JSON output. 

CI seems to be failing now for an unrelated reason:

> fatal: Remote branch V2 not found in upstream origin

pbennett:
ping

bwmx:
ping, this missing functionality is making it difficult for new users to get onboarded to algorand development, they get presented with typescript errors (that are difficult for a new user to resolve). It's undoubtedly frustrating for anyone else using the standard toolsets provided for development, unelegant hacks and workaround are the current solution.

pbennett:
@jasonpaulos  ?

jasonpaulos (approved):
Sorry for the delay on this. I was having trouble deciding between whether the new properties should be optional on `ABIMethod` and `ABIContract`, like they are in this PR, or if they should be non-optional `ARC28Event[]` and `boolean` types, which get coerced from undefined to a default value.

However, I now believe there is value in keeping the types optional, since a value of undefined can indicate that no information is available about these extension fields. For example, I think it's beneficial to be able to differentiate between a contract that claims it emits no events and one that claims nothing at all about events.

Thanks again @joe-p for this contribution

PR #857, opened 2024-02-20
Typed indexer responses

### LINKED ISSUE #847: Type indexer responses
#776 added typed responses for algod, and this issue is to do the same for indexer

jasonpaulos:
Adds types for indexer responses, similar to what we did for algod ones for v3 in #776

Fixes #847


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="V2"
+SDK_TESTING_BRANCH="update-indexer-mocks"
```
jasonpaulos, line 3:
```suggestion
SDK_TESTING_BRANCH="V2"
```
TODO: revert when https://github.com/algorand/algorand-sdk-testing/pull/308 is merged

tests/cucumber/steps/steps.js
```diff
@@ -2143,35 +2473,39 @@ module.exports = function getSteps(options) {
   When(
     'we make a Lookup Block call against round {int}',
     async function (round) {
-      await this.indexerClient.lookupBlock(round).do();
+      await doOrDoRaw(this.indexerClient.lookupBlock(round));
```
gmalouf, line 2476: The old syntax with a `do()` at the end looks nicer, but assume this was needed for a good reason.

gmalouf (approved):

tests/cucumber/steps/steps.js
jasonpaulos: Yes, this is necessary because we run some requests against a mock server that always returns an empty response with the intent to ensure that the right path is requested.

Now that indexer has typed responses, using the `do()` method will attempt to parse a response and fail. We have a `doRaw()` method which doesn't parse, which is ideal for this case, and the `doOrDoRaw` function choose the right one based on the specific test.

PR #860, opened 2024-03-07
feat: expose AtomicTransactionComposer.methodCalls

joe-p:
Makes `AtomicTransactionComposer.methodCalls` public. This is useful for modifying/retrieving the method calls in an ATC externally.

Specifically useful for https://github.com/algorandfoundation/algokit-utils-ts/pull/234

### DIFF
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -122,9 +122,11 @@ export class AtomicTransactionComposer {
   /** The maximum size of an atomic transaction group. */
   static MAX_GROUP_SIZE: number = 16;
 
+  /** A map of tranasction indexes to the respective ABI method being called. Used for properly decoding return values. */
+  methodCalls: Map<number, ABIMethod> = new Map();
+
   private status = AtomicTransactionComposerStatus.BUILDING;
   private transactions: TransactionWithSigner[] = [];
-  private methodCalls: Map<number, ABIMethod> = new Map();
   private signedTxns: Uint8Array[] = [];
   private txIDs: string[] = [];
 


src/composer.ts
```diff
@@ -122,9 +122,11 @@ export class AtomicTransactionComposer {
   /** The maximum size of an atomic transaction group. */
   static MAX_GROUP_SIZE: number = 16;
 
+  /** A map of tranasction indexes to the respective ABI method being called. Used for properly decoding return values. */
+  methodCalls: Map<number, ABIMethod> = new Map();
```
jasonpaulos, line 126: I'd prefer you leave the field as it was, but introduce a new getter method like `getMethodCalls` which returns a copy of the map.

That helps with maintainability and prevents clients from unintentionally modifying the internal map

joe-p: The problem is that modifying the map is sometimes necessary. For example, if you have an ABI method call that takes in another ABI method call as an argument. The ATC does not currently support this, so you need to get the `appl` transaction and update `methodCalls` if you want the return values. I suppose we could add support for that, but it would be a larger change to the ATC.

jasonpaulos: If that's the use case you're pursuing, then yes I think we need to talk about how the interface can support that. Directly modifying the internal map is not a stable solution

joe-p: I suppose we could also have a method `setMethodCall(idx: number, method: ABIMethod)` in addition to `getMethodCalls`

joe-p: And there can be sanity checks that it's an appl that doesn't already have a method set

robdmoore: Idea: what if we allow you to optionally pass in an ABIMethod when you call `addTransaction`? and then have a readonly property to expose method calls so you can read them out after using `addMethodCall`. Those two together would suffice?

Issue #846, opened 2023-12-25
BlockResponse class incorrectly modeled


SilentRhetoric:
### Subject of the issue

The BlockResponse.from_obj_for_encoding() returns a class object that has incorrect typing.

* Software version: 2.7.0

### Steps to reproduce

```
const blockRespRaw = algodClient.block(12345).do()

const incorrectTypedBlockResponse = modelsv2.BlockResponse.from_object_for_encoding(blockResponseRaw)
```

### Expected behaviour

Passing the `algodClient.block()` through `modelsv2.BlockResponse.from_obj_for_encoding()` should produce an object in which the `block` property is a `Block` and not a `BlockHeader`.

### Actual behaviour

Instead, the `block` property of the `BlockResponse` class is typed as a `BlockHeader` and thus does not allow accessing the rest of the block contents.

[https://github.com/algorand/js-algorand-sdk/blob/e947e3395bb074fb496fbdda91650137b5670127/src/client/v2/algod/models/types.ts#L1877](https://github.com/algorand/js-algorand-sdk/blob/e947e3395bb074fb496fbdda91650137b5670127/src/client/v2/algod/models/types.ts#L1877)

jasonpaulos:
Good point, this type is not accurate

jasonpaulos:
Fixed by #875

PR #862, opened 2024-03-25
3.0.0: Improve object encoding and decoding

jasonpaulos:
* Upgrade to the newly released `algorand-msgpack` library
* Remove the `from_obj_for_encoding` and `get_obj_for_encoding` methods. In their place, introduce the following:
  * An `Encodable` interface with a `toEncodingData` method, and a static `fromEncodingData` method. "Encoding data" is an intermediate representation of the object in a JS `Map`, similar to the output of our Python SDK's `dictify` method.
  * The `Encodable` interface also has a `getEncodingSchema` method and a static `encodingSchema` property, which return instances of the new `Schema` base class.
  * The `Schema` class is how "encoding data" gets transformed for encoding/decoding to/from JSON and msgpack. For example, `AddressSchema` converts an `Address` class to its 32-byte public key for encoding in msgpack, and for JSON it converts the address to its base32 string form. Specifically these translations happen in the `prepareMsgpack`/`fromPreparedMsgpack` and `prepareJSON`/`fromPreparedJSON` methods.
* Replace `EncodedSignedTransaction` interface with `SignedTransaction` class

While the introduction of `Schema` adds some complexity, it also has these benefits:
* Consolidating encoding decisions which otherwise must be spread across every class's encoding/decoding functions
* Reducing mistakes which can be made in each implementing class. No longer do they need to worry about omitting empty values or correctly dealing with base64 encoding for JSON. The `toEncodingData` and `fromEncodingData` can be more straightforward translations now.
* Real, configurable support for omitting empty values. The `NamedMapSchema` class determines if empty values should be omitted and seamlessly restores omitted values when decoding.
  * Each map field has a boolean property `omitEmpty` which, if true, omits default values during encoding.
  * When decoding, if a field has been omitted, sometimes we don't want to restore it to that field's default value, we'd rather just know if it wasn't present. For this reason there is the `OptionalSchema` which can wrap another schema. It will restore omitted values to `undefined`.
* Opens up the possibility of using non-string keys, which are necessary to support blocks and stateproof transactions (to be addressed in a later PR)

This relies on generator changes as well from https://github.com/algorand/generator/pull/73


jasonpaulos:
@emg110 this is still a work in progress, but yes, there are breaking change here. That's why this is going into the v3 branch

.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="V2"
+SDK_TESTING_BRANCH="fix-mock-response-txn-encoding-types"
```
jasonpaulos, line 3: TODO: revert once https://github.com/algorand/algorand-sdk-testing/pull/310 is merged
```suggestion
SDK_TESTING_BRANCH="V2"
```

src/client/v2/algod/models/types.ts
jasonpaulos, line 1: As a reminder, this file is generated, and its changes are a result of https://github.com/algorand/generator/pull/73

src/client/v2/indexer/models/types.ts
jasonpaulos, line 1: As a reminder, this file is generated, and its changes are a result of https://github.com/algorand/generator/pull/73

src/multisig.ts
jasonpaulos, line 1: These changes are mostly to move code which relies on `SignedTransaction` (i.e. code which produces signed msig txns) to `src/multisigSigning.ts`

src/encoding/schema/map.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (data.has(entry.key) && data.get(entry.key) !== undefined) {
+        if (
+          !entry.required ||
+          !entry.valueSchema.isDefaultValue(data.get(entry.key))
+        ) {
+          return false;
+        }
+      }
+    }
+    // TODO: examine if there are any extra keys?
```
gmalouf, line 45: how urgent is this?

src/encoding/schema/uint64.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ensureUint64 } from '../../utils/utils.js';
+
+/* eslint-disable class-methods-use-this */
+
+export class Uint64Schema extends Schema {
+  public defaultValue(): bigint {
+    return BigInt(0);
+  }
+
+  public isDefaultValue(data: unknown): boolean {
```
gmalouf, line 11: I recall some discussion around the integer type defaults, trying to recall what is impacted by that (not necessarily this function).

src/encoding/encoding.ts
```diff
@@ -10,7 +10,16 @@
  *  5. Binary blob should be used for binary data and string for strings
  *  */
 
-import * as msgpack from 'algo-msgpack-with-bigint';
+import {
+  encode as msgpackEncode,
```
gmalouf, line 14: where are encode/decode coming from here? Are these newly created somewhere (I may find later in review)

src/encoding/encoding.ts
```diff
@@ -63,5 +72,145 @@ export function encode(obj: Record<string | number | symbol, any>) {
...
+): T {
+  const decoded: JSONEncodingData = parseJSON(encoded, {
+    intDecoding: IntDecoding.MIXED,
+  });
+  return c.fromEncodingData(
+    c.encodingSchema.fromPreparedJSON(decoded) as JSONEncodingData
+  );
+}
+
+export function encodeJSON(e: Encodable, space?: string | number): string {
```
gmalouf, line 213: Where would the encode/decode JSON testing be found? (Do find some direct calls to the prepareJSON function off the Schema base class)

src/signing.ts
```diff
@@ -0,0 +1,95 @@
+import * as nacl from './nacl/naclWrappers.js';
+import { Address } from './encoding/address.js';
+import * as encoding from './encoding/encoding.js';
+import { SignedTransaction } from './signedTransaction.js';
+import { Transaction } from './transaction.js';
+import { LogicSig, LogicSigAccount } from './logicsig.js';
+import { addressFromMultisigPreImg } from './multisig.js';
+
+function signLogicSigTransactionWithAddress(
```
gmalouf, line 9: Not positive this is covered by testing (pre-existing).

src/boxStorage.ts
```diff
@@ -1,11 +1,10 @@
-import { EncodedBoxReference } from './types/transactions/index.js';
 import { BoxReference } from './types/transactions/base.js';
 
-function translateBoxReference(
+function boxReferenceToEncodingData(
   reference: BoxReference,
   foreignApps: bigint[],
   appIndex: bigint
-): EncodedBoxReference {
+): Map<string, unknown> {
```
gmalouf, line 7: What was rationale for not going with an alias for the types both here and on `translateBoxReference`?

src/signedTransaction.ts
```diff
@@ -0,0 +1,171 @@
...
+    {
+      key: 'txn',
+      valueSchema: Transaction.encodingSchema,
+      required: true,
+      omitEmpty: true,
+    },
+    {
+      key: 'sig',
+      valueSchema: new FixedLengthByteArraySchema(64),
+      required: false,
```
gmalouf, line 32: shouldn't sig be required on a signed transaction?

gmalouf: oh I guess it would be absent if msig present

src/signedTransaction.ts
```diff
@@ -0,0 +1,171 @@
...
+    this.sig = sig;
+    this.msig = msig;
+    this.lsig = lsig;
+    this.sgnr = sgnr;
+
+    let numberOfSigs = 0;
+    if (sig) numberOfSigs += 1;
+    if (msig) numberOfSigs += 1;
+    if (lsig) numberOfSigs += 1;
+    if (numberOfSigs > 1) {
```
gmalouf, line 103: okay this check makes me feel better

src/signedTransaction.ts
```diff
@@ -0,0 +1,171 @@
...
+      data.set('lsig', this.lsig.toEncodingData());
+    }
+    return data;
+  }
+
+  public static fromEncodingData(data: unknown): SignedTransaction {
+    if (!(data instanceof Map)) {
+      throw new Error(`Invalid decoded SignedTransaction: ${data}`);
+    }
+    return new SignedTransaction({
```
gmalouf, line 134: should there be a check for more than one signature on decode?

gmalouf (commented):
Left a few asks, overall makes sense and is definitely cleaner/seemingly more sustainable. Some concern about exported functions not covered by tests (even before these changes).

src/types/blockHeader.ts
```diff
@@ -79,3 +88,141 @@ export default interface BlockHeader {
...
+  header: BlockHeader
+): Map<string, unknown> {
+  return new Map<string, unknown>([
+    ['fees', header.fees],
+    ['frac', header.frac],
+    ['gen', header.gen],
+    ['gh', header.gh],
+    ['prev', header.prev],
+    ['proto', header.proto],
+    ['rate', header.rate],
```
algorandskiy, line 218: missing `TxnCounter uint64 `codec:"tc"``
missing `RewardsLevel uint64 `codec:"earn"`` from rewardState, not sure what is the rule here.
also missing `partupdrmv` and all `UpgradeState` fields except `proto`

src/signing.ts
algorandskiy: appears to be covered as signLogicSigTransaction -> signLogicSigTransactionObject -> signLogicSigTransactionObject but not sure if all cases are covered

src/signedTransaction.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  public toEncodingData(): Map<string, unknown> {
+    const data = new Map<string, unknown>([
+      ['txn', this.txn.toEncodingData()],
+      ['sig', this.sig],
+      ['sgnr', this.sgnr],
+    ]);
+    if (this.msig) {
+      data.set('msig', encodedMultiSigToEncodingData(this.msig));
```
algorandskiy, line 122: wouldn't it retain the empty `sig` property if msig provided?

src/multisigSigning.ts
```diff
@@ -0,0 +1,359 @@
...
+  signedTxn.msig!.subsig.forEach((subsig, i) => {
+    if (nacl.bytesEqual(subsig.pk, myPk)) {
+      keyExist = true;
+      signedTxn.msig!.subsig[i].s = rawSig;
+    }
+  });
+  if (!keyExist) {
+    throw new Error(MULTISIG_KEY_NOT_EXIST_ERROR_MSG);
+  }
+
```
algorandskiy, line 119: this misses msig check compare to the [original code](https://github.com/algorand/js-algorand-sdk/pull/862/files#diff-f860886128b9a9712adffc0a93540ef15ec333cedf3a2bc40953698001c74c86L233)
```
  const msigAddr = addressFromMultisigPreImg({
    version,
    threshold,
    pks,
  });
  const senderAddr = signedTxn.txn.snd
    ? new Address(signedTxn.txn.snd)
    : Address.zeroAddress();
  if (!senderAddr.equals(msigAddr)) {
    signedTxn.sgnr = msigAddr.publicKey;
  }
```

src/encoding/encoding.ts
jasonpaulos: These are from the `algorand-msgpack` library: https://github.com/algorand/msgpack-javascript/blob/main/src/index.ts

src/boxStorage.ts
jasonpaulos: This PR removes many named object types like `EncodedBoxReference` and `EncodedTransaction` which were used as the intermediate representation before JSON/msgpack encoding.

The reason being that we've upgraded from objects to maps to support non-string keys. And with maps, there's no convenient way to create a type alias. I actually think this change is good, as before the named object types were used in many places where they shouldn't have been, and their typed nature implied that certain fields were checked/correctly typed when in many cases we casted directly to those types after decoding with no additional checks.

src/signedTransaction.ts
jasonpaulos: I'm not so sure about this. If you try to create a signed txn with more than one signature, then yes we should error and stop you because you're doing something wrong and it's best to fail early.

But if you're just trying to read such a transaction that someone else created, I don't see a clear benefit to erroring in that case. There are almost an unlimited number of ways you can invalidate a transaction, and I'm not sure it's worth the effort to try to validate every field completely upon decoding (we do validate that field types are correct, but field values/consistency is another thing entirely). That will be done when you send the transaction to a node for processing anyway.

src/signedTransaction.ts
jasonpaulos: If `msig` is defined then `sig` must be `undefined`. Undefined values are [dropped during msgpack](https://github.com/algorand/js-algorand-sdk/blob/b864c24ac3007a2bba08f0bad1c98d1617891ebe/src/encoding/schema/map.ts#L62-L64) and JSON preparation, so it makes no difference.

A previous version of the code looked something like:
```javascript
if (this.sig) {
    data.set('sig', this.sig)
}
if (this.msig) {
    data.set('msig', encodedMultiSigToEncodingData(this.msig))
}
```
Due to dropping undefined values this is equivalent to the current code.

src/multisigSigning.ts
jasonpaulos: Yes, I removed this from `createMultisigTransactionWithSignature` because it's already checked in `createMultisigTransaction`, which is called at the beginning of the function.

`createMultisigTransaction`:
https://github.com/algorand/js-algorand-sdk/blob/b864c24ac3007a2bba08f0bad1c98d1617891ebe/src/multisigSigning.ts#L54-L64

src/boxStorage.ts
```diff
@@ -22,28 +21,24 @@ function translateBoxReference(
...
-  return encodedReference;
+  return new Map<string, number | Uint8Array>([
+    ['i', index],
+    ['n', referenceName],
+  ]);
 }
 
 /**
  * translateBoxReferences translates an array of BoxReferences with app IDs
  * into an array of EncodedBoxReferences with foreign indices.
```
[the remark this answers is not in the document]
jasonpaulos: Fixed in f79528e68ce60e0a1142fab4c67e6ebdb4379caf

src/encoding/schema/map.ts
jasonpaulos: I looked into it, and ignoring extra keys seems to make the most sense. Comment removed and unit tests with extra keys added in f79528e68ce60e0a1142fab4c67e6ebdb4379caf

src/client/v2/algod/accountApplicationInformation.ts
```diff
@@ -24,6 +24,8 @@ export default class AccountApplicationInformation extends JSONRequest<
 
   // eslint-disable-next-line class-methods-use-this
   prepare(body: Record<string, any>): AccountApplicationResponse {
-    return AccountApplicationResponse.from_obj_for_encoding(body);
+    return AccountApplicationResponse.fromEncodingData(
```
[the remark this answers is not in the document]
jasonpaulos: Yes this is pretty verbose and not very nice.

When decoding from a serialized JSON string, [`decodeJSON`](https://github.com/algorand/js-algorand-sdk/blob/f79528e68ce60e0a1142fab4c67e6ebdb4379caf/src/encoding/encoding.ts#L201) can be used, e.g. `decodeJSON(str, AccountApplicationResponse)`. This is what I expect users will need the vast majority of the time, not the verbose code present in the request classes.

In these request classes, we are in a bit of weird situation when `prepare` function is called, since the JSON response is already parsed into an object. I have a desire to change this so that the JSON string is the input to `prepare`, then `decodeJSON` can be used and the code would be cleaner. But this PR has already grown large, so I didn't want to attempt that now.

src/dryrun.ts
```diff
@@ -139,7 +139,7 @@ export async function createDryrun({
   await Promise.all(acctPromises);
 
   return new DryrunRequest({
-    txns: txns.map((st) => ({ ...st, txn: st.txn.get_obj_for_encoding() })),
```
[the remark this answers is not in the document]
jasonpaulos: Not really, that's more how the old code work, where `get_obj_for_encoding` was implemented generically by `BaseModel` and it attempted to call `get_obj_for_encoding` on all field values. See this file from develop: https://github.com/algorand/js-algorand-sdk/blob/981905ad8511cea706fc77d0c133bb1962f76927/src/client/v2/basemodel.ts

With the changes in this PR, it's more like `DryrunRequest` knows which of its fields are `Encodable` and it will call `toEncodingData` on them when its own `toEncodingData` method is called. There is no generic searching all objects for the magic function name.

src/encoding/schema/boolean.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/* eslint-disable class-methods-use-this */
+
+export class BooleanSchema extends Schema {
+  public defaultValue(): boolean {
+    return false;
+  }
+
+  public isDefaultValue(data: unknown): boolean {
+    return data === false;
```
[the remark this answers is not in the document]
jasonpaulos: `NamedMapSchema` is where omitting empty values is implemented, and it works without needing to allow undefined here.

Separately, I recently introduced an [`OptionalSchema`](https://github.com/algorand/js-algorand-sdk/blob/f79528e68ce60e0a1142fab4c67e6ebdb4379caf/src/encoding/schema/optional.ts#L6-L14) that you can wrap around any other schema to allow `undefined` values. It doesn't change how maps omit empty values, but it does allow for values to be decoded as `undefined`, which is useful in a language like JS without any native notion of zero values.

src/types/blockHeader.ts
jasonpaulos: Yes this block header representation is very outdated and wrong. We have an issue to fix in #846, and since this PR is large I didn't want to address that now. I only preserved the existing structure, even though it needs improvement

src/signing.ts
jasonpaulos: This is an internal function, so it's not tested directly. But as Pavel pointed out, since every call to `signLogicSigTransaction` and `signLogicSigTransactionObject` resolves to this, and those are covered well [here](https://github.com/algorand/js-algorand-sdk/blob/6c90132e5c9358ea2e4010c4eadf9e4e12527190/tests/8.LogicSig.ts#L440-L647), this should have good coverage.

src/encoding/schema/uint64.ts
jasonpaulos: I think you may be referring to whether integers get decoded as numbers vs bigints?

This function is concerning which values get omitted from the msgpack/JSON encodings. We check for both numbers and bigints to cover all cases.

gmalouf (dismissed):

algorandskiy (approved):

gmalouf (approved):

PR #874, opened 2024-06-09
readonly field on ABIMethodParams

scholtz:
generated client throws error on every readonly method with missing readonly field on ABIMethodParams


joe-p:
Duplicate of #856. Will fix that PR tomorrow

PR #875, opened 2024-06-11
3.0.0: Correctly model blocks

### LINKED ISSUE #846: BlockResponse class incorrectly modeled

jasonpaulos:
Update this SDK with a correct view of `BlockHeader`, and a `Block` response which includes the transaction payset.

Fixes #846


src/client/v2/algod/models/types.ts
jasonpaulos, line 1: These generated changes are a result of https://github.com/algorand/generator/pull/78

gmalouf (approved):

PR #876, opened 2024-06-21
Fix stateproof txn representation

### LINKED ISSUE #855: Fix incorrect stateproof representation
Currently the stateproof transaction fields are incorrect: https://github.com/algorand/js-algorand-sdk/blob/6fdfa116f0e47e68982281028fdc817577c6aab7/src/transaction.ts#L209-L211

`stateProof` and `stateProofMessage` are not byte arrays, the are objects.

Additionally, the block header type contains an invalid representation of state proof tracking info: https://github.com/algorand/js-algorand-sdk/blob/6fdfa116f0e47e68982281028fdc817577c6aab7/src/types/blockHeader.ts#L80

It is not possible to decode a `Map` object from JSON or msgpack, and the value type should be an object, not a byte array.

It remains to be seen whether the best approach is to fully type stateproof objects, or reference then generically, but something must be done to improve this.

Additionally, the stateproof types use integers as map keys, so it will be difficult to get these to encode and decode properly.

jasonpaulos:
Corrects the representation of a state proof transaction in this SDK.

Fixes #855


package.json
jasonpaulos, line 1: Not strictly necessary, but I upgraded these dependencies while debugging and decided to keep them in this PR

gmalouf (approved):

PR #878, opened 2024-07-22
3.0.0: Support special case raw binary strings

### LINKED ISSUE #724: Incorrect msgpack decoding on global state delta values
### Subject of the issue

When requesting a Block from algod and decoding each transaction, the global state deltas of the `apply_delta.eval_delta` contain incorrect data.

using 

```ts
   const block = await client.block(block_number).do();
   // ...
  block.block.txns.forEach((t) => {
    if ("dt" in t) {
      if ("gd" in t["dt"]) {
        for (const k of Object.keys(t["dt"]["gd"])) {
          console.log(t["dt"]["gd"][k]);
        }
      }
    }
  });
 ```

logs a string like 
```
'Ibf\x88³\x13|ù򦞴GÁ֨쌥g\x00װ§࠲\x93ش\x00\x00\x00\x00\x01᫰\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00.Y«A\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00ÿÿþ°ݠ\x8Fက\x01O"\x9Fp\x1E\x00\x00\x03^\x04쪅\x00\x00\x00\x92I\x7Fû\x8A\x00\x00\x00\x05ő\x04뀀\x00\x01¾(\x05 \x00\x00\x00\x01
```

but expecting 
```
'Ibf\x88\xb3\x13|\xf9\xf6&\x9e\xf4G\xc3\x81\xe0\x96(\xec\xcc\xa5g\x00\xd7\xf0\xa7\xe0\xe0\xf2\x93\xd8t\x00\x00\x00\x00\x01\xe1\xabp\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00.Y\xabA\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\xfe\xb0\xdd`\x8f\xe1\x00\x00\x01O"\x9fp\x1e\x00\x00\x03^\x04\xec\xea\x05\x00\x00\x00\x92I\x7f\xfb\x8a\x00\x00\x00\x05\xc5\x11\x04\xeb\x00\x00\x00\x01\xbe(\x05\xe2\x00\x00\x00\x00\x00\x00'
```

This seems to be happing in the `msgpack.decode` call since even with the raw bytes read from disk the same result appears.

Using `msgpack-lite` locally and setting the codec `useraw` field, the data is decoded correctly:
```ts
   let rawBlock = await client.block(block_number).doRaw();
   const block = mpk.decode(rawBlock, {
     codec: mpk.createCodec({ useraw: true }),
   }) as Record<string, any>;

```

jasonpaulos:
Some fields in `EvalDelta` are encoded as msgpack strings, but they actually contain arbitrary binary data. In order to properly decode these values in this SDK, this PR uses raw string decoding made available from our forked https://github.com/algorand/msgpack-javascript

This PR relies on unreleased changes to the msgpack library, made in this PR: https://github.com/algorand/msgpack-javascript/pull/6. That should be merged and released before this is mergable.

Closes #724


.circleci/config.yml
```diff
@@ -138,9 +138,21 @@ commands:
...
+            git checkout raw-string-encoding
+            npm i
+            npm run prepare
+            npm pack
       - run:
           name: npm ci
           command: |
             set -e
             npm ci
+            cp ../msgpack-javascript/algorand-msgpack-1.0.1.tgz algorand-msgpack-1.0.1.tgz
```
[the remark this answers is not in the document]
gmalouf: Okay, so we should not merge with this as-is.

src/group.ts
```diff
@@ -12,7 +12,7 @@ function txGroupPreimage(txnHashes: Uint8Array[]): Uint8Array {
       `${txnHashes.length} transactions grouped together but max group size is ${ALGORAND_MAX_TX_GROUP_SIZE}`
     );
   }
-  const bytes = encoding.encode({
+  const bytes = encodeObj({
```
algorandskiy, line 15: isn't it deprecated now so that a replacement should be used?

src/encoding/schema/map.ts
```diff
@@ -424,3 +449,129 @@ export class StringMapSchema extends Schema {
...
+      string | number | bigint | Uint8Array,
+      MsgpackEncodingData
+    >();
+    for (const [key, val] of value) {
+      newMap.set(
+        removeRawStringsFromMsgpackValues(key) as
+          | string
+          | number
+          | bigint
+          | Uint8Array,
```
algorandskiy, line 470: what does this `as | string | number` syntax mean?

src/encoding/schema/map.ts
```diff
@@ -424,3 +449,129 @@ export class StringMapSchema extends Schema {
     return map;
   }
 }
+
+function removeRawStringsFromMsgpackValues(
```
algorandskiy, line 453: not sure I understand "remove" part in its name - isn't it converts raw strings instances to bytes?

algorandskiy (commented):
The code LGTM. Not sure about deprecation stuff.

src/encoding/schema/map.ts
jasonpaulos: The `as X` syntax is a type cast, and `string | number | bigint | Uint8Array` is the union type of all of these individual types. This cast is needed because the return type of the function is the more general `MsgpackEncodingData`, but map keys can only be a subset of the `MsgpackEncodingData` definition.

src/encoding/schema/map.ts
jasonpaulos: Right, this is not the best name. Updated in 32c59407

src/group.ts
jasonpaulos: Good point, updated in 32c59407

src/encoding/schema/map.ts
```diff
@@ -424,3 +449,136 @@ export class StringMapSchema extends Schema {
     return map;
   }
 }
+
+/**
+ * Converts any RawBinaryString values to regular strings in a MsgpackEncodingData object.
+ *
+ * Note this conversion may be lossy if the binary data is not valid UTF-8.
```
gmalouf, line 456: Should there be an parameter option to error rather than silently be lossy?

v2_TO_v3_MIGRATION_GUIDE.md
```diff
@@ -301,6 +301,15 @@ In order to facilitate `bigint` as a first-class type in this SDK, additional JS
 
 If your v2 code uses `JSON.parse` or `JSON.stringify` on types which can now contain `bigint`s in v3, you may receive an error such as `TypeError: Do not know how to serialize a BigInt`. Consider using these new functions instead. Or, if the types are `Encodable`, use the new `encodeJSON` and `decodeJSON` functions described in the [Object Encoding and Decoding](#object-encoding-and-decoding) section.
 
+### Msgpack Operations
+
+The functions `encodeObj` and `decodeObj`, used to encode and decode msgpack objects, have been deprecated in v3 in favor of new functions, `msgpackRawEncode` and `msgpackRawDecode`. These functions have clearer names and differ slightly from the old functions. Specifically:
```
gmalouf, line 306: These functions are being renamed in this PR, just noting that if we land on not renaming / deprecating simultaneously that this documentation needs to be updated.

gmalouf (commented):
Overall is looking good, a few comments and questions to go through.

src/encoding/encoding.ts
```diff
@@ -43,44 +44,115 @@ function containsEmpty(obj: Record<string | number | symbol, any>) {
...
- * encode encodes objects using msgpack
- * @param obj - a dictionary to be encoded. Must not contain empty or 0 values.
- * @returns msgpack representation of the object
+ * encodeObj takes a javascript object and returns its msgpack encoding
+ * Note that the encoding sorts the fields alphabetically
+ * @param o - js object to be encoded. Must not contain empty or 0 values.
+ * @returns Uint8Array binary representation
  * @throws Error containing ERROR_CONTAINS_EMPTY_STRING if the object contains empty or zero values
+ *
+ * @deprecated Use {@link msgpackRawEncode} instead. Note that function does not
```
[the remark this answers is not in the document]
jasonpaulos: The rename is not externally visible, it's only to clear up an inconsistency. This function has always been exported externally as `encodeObj`, and I found the fact that the internal and external names differed to be an unnecessary complexity.

Externally there's no difference here other than the deprecation.

src/encoding/encoding.ts
```diff
@@ -43,44 +44,115 @@ function containsEmpty(obj: Record<string | number | symbol, any>) {
...
-export function decodeAsMap(encoded: ArrayLike<number>) {
-  // TODO: consider different int mode
-  const options: DecoderOptions = { intMode: IntMode.MIXED, useMap: true };
-  return msgpackDecode(encoded, options);
+/**
+ * decodeObj takes a Uint8Array and returns its javascript obj
+ * @param o - Uint8Array to decode
+ * @returns object
+ *
+ * @deprecated Use {@link msgpackRawDecode} instead. Note that this function uses `IntDecoding.MIXED`
```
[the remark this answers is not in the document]
jasonpaulos: Agree 100%, and it's the same situation as `encodeObj` here as well

src/encoding/schema/binarystring.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * SpecialCaseBinaryStringSchema is a schema for byte arrays which are encoded
+ * as strings in msgpack and JSON.
+ *
+ * This schema allows lossless conversion between the in memory representation
+ * and the msgpack encoded representation, but NOT between the in memory and
+ * JSON encoded representations if the byte array contains invalid UTF-8
+ * sequences.
+ */
+export class SpecialCaseBinaryStringSchema extends Schema {
```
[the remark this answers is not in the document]
jasonpaulos: In a msgpack encoded block, some fields from the eval delta are encoded as msgpack strings when they actually contain arbitrary bytes. The affected fields are eval delta logs and global and local state delta keys and values. I have updated the `EvalDelta` and `ValueDelta` classes from `src/types/block.ts` to use this new schema and the new `SpecialCaseBinaryStringMapSchema`.

I go into more detail about the underlying problem in this issue: https://github.com/algorand/go-algorand/issues/5807#issuecomment-1781599209

Unless we introduce more fields into blocks which suffer from this same drawback (which I don't think we want to and we've taken steps to prevent), there should be no further need for these schemas in new fields/classes.

src/encoding/schema/map.ts
jasonpaulos: Yes, after more consideration I think that's the safest thing to do. Working on that now

jasonpaulos: Implemented in 561aa4db

gmalouf (approved):

PR #881, opened 2024-08-07
REST API TEAL bytes fix

jasonpaulos:
Algod types TEAL byte keys and byte values as strings and base64 encodes them in responses. There was some discussion in https://github.com/algorand/go-algorand/pull/5492 about changing the types on algod, but we decided against it.

This PR updates the types client side so that users don't need to manually decode base64 strings.

A few byte conversion operations were cleaned up as well in the example files.

Relies on https://github.com/algorand/generator/pull/79 to update the generated REST models.

### DIFF
--- a/src/client/v2/algod/models/types.ts
+++ b/src/client/v2/algod/models/types.ts
@@ -558,7 +558,7 @@ export class Account implements Encodable {
 
   static fromEncodingData(data: unknown): Account {
     if (!(data instanceof Map)) {
-      throw new Error(`Invalid decoded logic sig account: ${data}`);
+      throw new Error(`Invalid decoded Account: ${data}`);
     }
     return new Account({
       address: data.get('address'),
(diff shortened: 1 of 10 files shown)


src/client/v2/algod/models/types.ts
```diff
@@ -7954,14 +7968,14 @@ export class TealKeyValue implements Encodable {
...
       (this.encodingSchemaValue as NamedMapSchema).pushEntries(
-        { key: 'key', valueSchema: new StringSchema(), omitEmpty: true },
+        { key: 'key', valueSchema: new ByteArraySchema(), omitEmpty: true },
         { key: 'value', valueSchema: TealValue.encodingSchema, omitEmpty: true }
       );
     }
     return this.encodingSchemaValue;
   }
 
-  public key: string;
```
jasonpaulos, line 7964: Note the only changes not related to the error string being updated are in `TealKeyValue` and `TealValue`

algorandskiy (approved):

PR #882, opened 2024-08-08
3.0.0: Fix remaining REST untyped responses

jasonpaulos:
Prior to this PR, the following algod request classes still had untyped responses: `Genesis`, `GetLedgerStateDelta`, `GetLedgerStateDeltaForTransactionGroup`, `GetTransactionGroupLedgerStateDeltasForRound`, `HealthCheck`, `Ready`, `SetBlockOffsetTimestamp`, `SetSyncRound`, `UnsetSyncRound`. These mostly fall into two categories: either they return no information, or they returned something related to state deltas and were passed over when #776 was merged.

This PR resolves this and fixes a few other related issues/improvements:
* `HTTPClient` no longer parses JSON responses. This simplifies the `prepare` method in request classes and no longer requires passing in JSON parsing options at the time of every request.
* POST & DELETE requests used to override the `do` method to implement their requests. That meant the `doRaw` method wasn't overridden and would produce incorrect results. Fixed by moving the override to a new `executeRequest` method which is called from both `do` and `doRaw`.
* Removed the second, unneeded generic type argument from `JSONRequest`
* Added `BlockHashSchema` because JSON block hashes get b32 encoded and prefixed with `blk-`.
* Added `LedgerStateDelta` and required subtypes. This comes with the caveat that the SDK does not currently support decoding maps with object keys (needed to support the `Txleases` field), so I've left that as an `UntypedValue` for now. I'll open a separate issue to track that, but I don't view it as urgent/required for v3.

Relies on https://github.com/algorand/generator/pull/80


algorandskiy (commented):
Changes look good to me.
It was a great effort to support all these types spawn by StateDelta but I do think it was mistake making StateDelta available through REST API. It was always thought as ledger internal so that easily modifiable as needed.

src/client/v2/algod/healthCheck.ts
```diff
@@ -1,25 +1,15 @@
...
  * healthCheck returns an empty object iff the node is running
  */
-export default class HealthCheck extends JSONRequest {
+export default class HealthCheck extends JSONRequest<void> {
   // eslint-disable-next-line class-methods-use-this
   path() {
     return '/health';
   }
 
-  async do(headers = {}) {
```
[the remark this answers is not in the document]
jasonpaulos: I believe the reason for it was so that an error could be thrown if the request failed. But it wasn't really effective, since if `res.ok` was false the underlying `c.get` method would already throw an error.

And since this method doesn't return any data, the bigint decoding doesn't really make a difference.

src/client/client.ts
```diff
@@ -14,12 +14,60 @@ interface ErrorWithAdditionalInfo extends Error {
...
+    this.format = options.format;
+    this.headers = options.headers;
+    this.status = options.status;
+    this.ok = options.ok;
+  }
+
+  /**
+   * Returns the response body as a string, ready to be parsed as JSON.
+   */
+  getJSONText(): string {
```
gmalouf, line 50: NIT: Why have the word JSON in the funcion name at all unless checking the content type?

src/client/v2/algod/getLedgerStateDeltaForTransactionGroup.ts
```diff
@@ -1,17 +1,24 @@
...
 
-export default class GetLedgerStateDeltaForTransactionGroup extends JSONRequest {
+export default class GetLedgerStateDeltaForTransactionGroup extends JSONRequest<LedgerStateDelta> {
   constructor(
     c: HTTPClient,
     private id: string
   ) {
     super(c);
-    this.query = { format: 'json' };
+    this.query = { format: 'msgpack' };
```
gmalouf, line 12: Noting the change from json here...

gmalouf (commented):
About 70% through, minimal comments so far.

src/client/client.ts
jasonpaulos: `this.text` is only populated if the response is JSON. This happens on line 232

src/client/v2/algod/getLedgerStateDeltaForTransactionGroup.ts
jasonpaulos: The main reason for this is that LedgerStateDeltas contain strings which are actually binary data. We can properly decode these from msgpack using the recently added special case binary string schemas. From JSON it's impossible to decode these properly, as the binary strings have already been converted to UTF-8.

tests/2.Encoding.ts
```diff
@@ -2622,4 +2738,481 @@ describe('encoding', () => {
       assert.deepStrictEqual(reencoded, encodedEvalDelta);
     });
   });
+  describe('LedgerStateDelta', () => {
```
gmalouf, line 2741: Needed to see this, crazy the amount of nested container objects required.

gmalouf (approved):

PR #883, opened 2024-08-12
3.0.0: Add ability to pass through fetch options

### LINKED ISSUE #858: Add ability to pass in fetch options for urlTokenBaseHTTPClient requests
## Problem

There is no way to abort the execution of a `fetch` request mid-flight. There is limited access to the fetch options and it is currently not possible to set the `signal` key to an `AbortSignal` from an `AbortController`.

- https://developer.mozilla.org/en-US/docs/Web/API/AbortController

## Solution

Allow for all additional options to be passed to the fetch client. This will also solve any future use cases that rely on access to the `fetch` options.

```javascript
client.getSomeRequest().do( headers, {signal} )
```

## Urgency

Very low

jasonpaulos:
When making requests, exposes an additional options object that can be passed through to the underlying `BaseHTTPClient`. For the default `URLTokenBaseHTTPClient`, this is treated as additional options to pass to the `fetch` network call.

Fixes #858

### DIFF
--- a/src/client/baseHTTPClient.ts
+++ b/src/client/baseHTTPClient.ts
@@ -26,8 +26,11 @@ export interface BaseHTTPClientError {
  * wallets to provide access to paid API services without leaking
  * the secret tokens/URLs.
  *
- * Note that post and delete also have an optional query parameter
- * This is to allow future extension where post and delete may have queries
+ * The parameter `customOptions` is an object that can be used to configure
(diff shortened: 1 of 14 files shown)


gmalouf (approved):

algorandskiy (approved):
The code looks right to me but was it tested somehow?

jasonpaulos:
Good point. I tested locally but it's always better to have a unit test. Added in 28ad187a

PR #892, opened 2024-08-21
Release PR Generation: Use prerelease tag in changelog generation

jasonpaulos:
Use the prerelease tag when generating the changelog for release PRs.

If a prerelease version is specified, right now that is not used when generating the release PR with the planned version tag.

For example, that causes this statement to be at the end of the changelog:

> **Full Changelog**: https://github.com/algorand/js-algorand-sdk/compare/v2.9.0...v3.0.0

Instead of:

> **Full Changelog**: https://github.com/algorand/js-algorand-sdk/compare/v2.9.0...v3.0.0-beta.1

I manually updated it in #891, but it would be nice to fix this at the source

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -81,6 +81,7 @@ jobs:
         id: build-changelog
         env:
           PREVIOUS_VERSION: ${{ steps.get-release.outputs.latest-tag }}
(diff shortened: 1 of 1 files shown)


onetechnical (approved):
I guess there's some hokeyness when handling beta releases, since they aren't exactly new releases; I suppose new versions of beta will still include previous beta changes. Overall I think this is fine, but may involve manual work to clean up follow up beta releases.

PR #911, opened 2024-12-13
feat: add Address as an ABIValue and handle encoding an Address

joe-p:
Allows `Address` to be an acceptable ABIValue and properly handles `Address` in `ABIAddressType.encode`

### DIFF
--- a/src/abi/abi_type.ts
+++ b/src/abi/abi_type.ts
@@ -13,7 +13,7 @@
     // | string
     // | (T1, ..., Tn)
 */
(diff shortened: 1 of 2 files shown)


gmalouf:
@joe-p can you add test coverage for this?

joe-p:
Added test case in [f9dceaf](https://github.com/algorand/js-algorand-sdk/pull/911/commits/f9dceaf6e83f079267c7824ca983383e3c88cafd). See header comment for implementation rationale. 

Also rebased against develop

gmalouf (approved):

PR #917, opened 2025-01-11
api: port regenerated models

algorandskiy:
Regenerate models and pick required changed: BlockHeaderResponse + Hearbeat.
Regenerating everything produced a [huge diff](https://github.com/algorand/js-algorand-sdk/pull/916) due to formatter (not sure why) and some pre-checks removal.


gmalouf (commented):
Are we skipping support for the cross SDK tests (may have to explicitly disable if so)?

gmalouf (approved):
Are we skipping support for the cross SDK tests (may have to explicitly disable if so)?

algorandskiy:
I'm working on the followup PR with deserialization code and tests.

PR #918, opened 2025-01-14
Incentives: Support heartbeat transaction (v2.9)

algorandskiy:
## Summary

Add heartbeat txn support. Note, this change includes models regenerated in https://github.com/algorand/js-algorand-sdk/pull/917 (i.e. subsequent PR)

## Testing

1. Added golden txn test
2. Added cross sdk tests

### DIFF
--- a/tests/cucumber/unit.tags
+++ b/tests/cucumber/unit.tags
@@ -2,6 +2,7 @@
 @unit.abijson.byname
 @unit.algod
 @unit.algod.ledger_refactoring
+@unit.algod.heartbeat.msgp
 @unit.applications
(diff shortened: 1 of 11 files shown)


tests/cucumber/unit.tags
```diff
@@ -2,6 +2,7 @@
 @unit.abijson.byname
 @unit.algod
 @unit.algod.ledger_refactoring
+@unit.algod.heartbeat.msgp
```
gmalouf, line 5: should we explicitly enable the json example here (I think it runs either way)?

gmalouf (approved):

PR #950, opened 2025-03-27
Devops: Release process update

gmalouf:
With this pull request (and changes already made to the SDK's settings), the release process for the Javascript SDK will be as follows:

1. Run create-release-pr.yml, which creates a release pull request.
2. Permissioned user approves the pull request (does not merge themselves)
3. Run release.yml, specifying the PR number output from (1) and the release version (tag) as inputs.
4. Doublecheck releases page, proper tag created, branch, PR merged/branch deleted, etc.

As a part of this, we moved away from `Git Flow` to a single mainline branch (named main).


.github/workflows/create-release-pr.yml
```diff
@@ -130,47 +141,25 @@ jobs:
         with:
           message: "bump up version to ${{ env.RELEASE_TAG }}"
 
-      - name: Create Pull Request to Master
+      - name: Create Pull Request to ${{ inputs.base_branch }}
+        if: ${{ env.PRE_RELEASE_VERSION == '' }}
```
onetechnical, line 145: If we don't create a PR for pre_release_version, when/how do we use pre_release_version?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+  workflow_dispatch:
+    inputs:
+      tag_version:
+        description: 'The tag_version, e.g. vx.x.x'
+        default: 'vx.x.x'
+        required: true
+        type: 'string'
+      node_version:
+        description: 'The node_version to build with.'
+        default: '22.x'
```
onetechnical, line 13: Same as the create PR, shouldn't this be:

```suggestion
        default: '22'
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
...
+        type: 'string'
+      dry_run:
+        description: 'Whether this is a dry_run publish.'
+        default: 'true'
+        required: true
+        type: 'boolean'
+
+jobs:
+  build_and_publish:
+    runs-on: ubuntu-latest
```
onetechnical, line 24: We should probably pin this to latest lts.

```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,163 @@
...
+      dry_run:
+        description: "Set to true to 'dry-run'/not actually publish to NPM."
+        default: "false"
+        required: true
+        type: boolean
+      node_version:
+        description: "The Node.js version to use for publishing"
+        required: true
+        type: string
+        default: "22.x"
```
onetechnical, line 23:
```suggestion
        default: "22"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,163 @@
...
+
+jobs:
+  release_and_publish:
+    runs-on: ubuntu-latest
```
onetechnical, line 27:
```suggestion
    runs-on: ubuntu-24.04
```

onetechnical (changes requested):
Had some comments/questions.

We also had a side discussion about "beta" push support. The way this works now I don't think it'll create release notes for a pre_release_version. Additionally, for publishing to npm, it would be nice to enable pushing beta versions for pre_release_versions, e.g. for a consensus upgrade we might want people to test on betanet before it hits mainnet. not clear if we actually need this support or not, though.

onetechnical:
Also possibly orthogonal, but since we're bumping node elsewhere, we probably also want to update it here:
https://github.com/algorand/js-algorand-sdk/blob/4742a84bc8af2f333a228f996464c7ec607ac0d4/.circleci/config.yml#L140

.github/workflows/create-release-pr.yml
```diff
@@ -8,10 +8,20 @@ on:
...
+        default: 'main'
+        required: true
+        type: string
       pre_release_version:
         description: "(Optional) Pre-Release version, e.g. 'beta.1'. Used mainly to support consensus release on betanet."
         required: false
         type: string
+      node_version:
+        description: 'The node_version to build with.'
+        default: '22.x'
```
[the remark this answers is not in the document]
gmalouf: I'll test it!

.github/workflows/create-release-pr.yml
gmalouf: Removed all pre-release support in latest update

onetechnical (approved):

PR #979, opened 2025-05-21
API: Algorand 4.1 api updates

gmalouf:


tests/cucumber/steps/steps.js
```diff
@@ -2157,9 +2157,20 @@ module.exports = function getSteps(options) {
...
   When(
-    'we make a GetApplicationBoxes call for applicationID {int} with max {int}',
-    async function (index, limit) {
-      await this.v2Client.getApplicationBoxes(index).max(limit).doRaw();
+    'we make a GetApplicationBoxes call for applicationID {int} with max {int} prefix {string} next {string} values {string}',
+    async function (index, limit, prefix, next, valuesAsString) {
+      let values = false;
+      if (valuesAsString === 'true') {
+        values = true;
+      }
```
jannotti, line 2165:
```suggestion
      const values = (valuesAsString === 'true')
```

jannotti (approved):
Looks good.

tests/cucumber/steps/steps.js
algorandskiy: +1 here and below

algorandskiy (approved):

tests/cucumber/steps/steps.js
gmalouf: Thanks, addressed both places in latest push.

jannotti (approved):

PR #985, opened 2025-06-02
CICD: implement tests in GitHub Actions

onetechnical:
## Summary

This is an implementation of unit and integration tests in GitHub Actions. Google Chrome and Firefox appear to already be installed in the ubuntu-24.04 image, so no need to install them here.

Chromedriver requires an explanation. Because we have it as a devDependency, the value for it can be overridden by what is in package.json. Because this is just for testing, We install a version of chromedriver that matches the installed google-chrome.

Oddly geckodriver doesn't appear to show any compatibility issues, but if it does later, we may have to revisit. (The driver version does not match the browser version, so we may have to figure out the right versioning.)

Some other changes:
- Removed interactive mode from dockerfile, as it was unnecessary
- Created a ci-test target that does not build a docker container, as it was redundant
- Upgraded chromedriver in package.json, because it was getting 404 errors in builds
- Updated tar-fs from 3.0.8 -> 3.0.9

## Testing

Ensure tests pass on this PR (including CircleCI).

The generate-docs job was tested on a fork.

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -36,7 +36,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ env.NODE_VERSION }}
-          cache: 'npm'      
+          cache: 'npm'
       - name: Install npm dependencies
         run: npm ci
       - name: Check Prettier formatting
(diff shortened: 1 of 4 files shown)


onetechnical:
I want to check if firefox is pre-installed, in which case I can skip a step.

.github/workflows/ci-pr.yml
```diff
@@ -97,11 +97,6 @@ jobs:
           # Verify the installation
           echo "Installed ChromeDriver version:"
           npx chromedriver --version
-      - name: Setup Firefox browser
```
gmalouf: Do the firefox tests still run?

PR #989, opened 2025-06-17
fix: add webpackloader to add Symbol.hasInstance

lempira:


scripts/webpackLoaderSymbolHasinstance.js
```diff
@@ -0,0 +1,78 @@
+function injectConstructorMarker(source, className, markerProperty) {
+  const constructorMarker = `
+    // @ts-ignore - Add unique identifier for cross-module instanceof compatibility
```
lempira, line 3: I had to add some `ts-ignore` to avoid tsc errors for the hasInstance  and _isAlgosdkAddress no being part of the type def.

PR #990, opened 2025-06-23
Fix/dual package hazard symbol hasinstance

lempira:
# Fix dual package hazard with Symbol.hasInstance

## Problem
When a package is loaded both as CommonJS and ES modules, `instanceof` checks can fail across module boundaries. This happens because the same class loaded through different module systems creates distinct class objects, causing `instanceof` to return false even for valid instances.

@neilcampbell created a repo to reproduce this error. I tested this fix against that repo, and it worked. 

## Solution
Added `Symbol.hasInstance` static methods to core classes along with unique property markers to properly handle instance checks across module boundaries. This follows the Node.js recommended pattern for handling dual package hazard.

For example:
```typescript
class Address {
  private readonly _isAlgosdkAddress = true;
  
  static [Symbol.hasInstance](instance: any): boolean {
    return !!(instance && instance._isAlgosdkAddress === true);
  }
}
```

This ensures that `instanceof` checks work correctly regardless of how the class is loaded (CommonJS or ESM).

We investigate another solution of trying to update the [webpack loader in this branch](https://github.com/lempira/js-algorand-sdk/tree/fix/instanceof) but that changed required an update to build process and thought this change would be easier. 

## Classes Updated
**Core Classes:**
- Address, Transaction, SignedTransaction, LogicSig, LogicSigAccount, AtomicTransactionComposer, ABIContract

**ABI Type Classes:**
- ABIUintType, ABIUfixedType, ABIAddressType, ABIBoolType, ABIByteType, ABIStringType, ABIArrayStaticType, ABIArrayDynamicType, ABITupleType

## Testing
Added comprehensive test suite in `tests/11.DualPackageHazard.ts` to verify cross-module instanceof behavior works correctly.

## Related
- Original issue https://github.com/algorandfoundation/algokit-utils-ts/issues/399
- Related draft PR https://github.com/algorand/js-algorand-sdk/pull/982


lempira:
> This means that any class that does NOT implement this fix may lead to unexpected when users try to use `instanceof` across exports. I don't think we need to do this for everything, but might be worth noting this in the documentation/README.
> 
> @lempira I think we'll probably want to also implement this for all the ABI types since they may get passed in to the SDK/algokit from users from different libs.

Updated to add the ABIType Classes

.eslintrc.js
```diff
@@ -41,6 +41,29 @@ module.exports = {
...
+          '_isAlgosdkABIAddressType',
+          '_isAlgosdkABIBoolType',
+          '_isAlgosdkABIByteType',
+          '_isAlgosdkABIStringType',
+          '_isAlgosdkABIArrayStaticType',
+          '_isAlgosdkABIArrayDynamicType',
+          '_isAlgosdkABITupleType',
+        ],
+      },
+    ],
```
lempira, line 66: I could just allow all leading `_` here. Is that ok for this repo?

joe-p:
@lempira I made PR that I think should fix the errors we see in the browser tests: https://github.com/lempira/js-algorand-sdk/pull/2

The test uses the same variable for uint and ufixed so I think TypeScript wasn't inferring it correctly.

PR #1000, opened 2025-09-10
types: add txn access

algorandskiy:
## Summary

Add `txn.Access` support similarly to other SDKs and tests.
The caveat is JS SDK uses the same types for deflated and inflated representation of box references (and so that holdings and locals inside resource reference) and accepts weakly typed addresses (`string`, `Address`).
This type acrobatics while maintaining proper inflated/deflated versions contributed to the PR size.

### DIFF
--- a/src/appAccess.ts
+++ b/src/appAccess.ts
@@ -254,7 +254,7 @@ export function foreignArraysToResourceReferences({
   function ensureApp(app: number | bigint) {
     let appFound = false;
     for (const rr of accessList) {
-      if (rr.appIndex && rr.appIndex === app) {
+      if (rr.appIndex === app) {
         appFound = true;
         break;
       }
--- a/src/appAccess.ts
+++ b/src/appAccess.ts
@@ -5,6 +5,7 @@ import {
   LocalsReference,
   ResourceReference,
 } from './types/transactions/base.js';
+import { ensureSafeUnsignedInteger } from './utils/utils.js';
 
 /**
  * resourceReferencesToEncodingData translates an array of ResourceReferences into an array of encoding data
@@ -40,7 +41,6 @@ export function resourceReferencesToEncodingData(
   }
 
   const zeroAddr = Address.zeroAddress();
-
   for (const rr of references) {
     if (rr.address || rr.assetIndex || rr.appIndex) {
       ensure(rr);
@@ -87,8 +87,8 @@ export function resourceReferencesToEncodingData(
         ])
       );
     }
-    if (rr.boxes) {
-      const b = rr.boxes;
+    if (rr.box) {
+      const b = rr.box;
       let appIdx = 0;
       if (b.appIndex && BigInt(b.appIndex) !== appIndex) {
         appIdx = ensure({ appIndex: b.appIndex });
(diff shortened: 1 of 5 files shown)


src/makeTxn.ts
```diff
@@ -408,10 +414,38 @@ export function makeApplicationCallTxnFromObject({
   lease,
   rekeyTo,
   suggestedParams,
-}: ApplicationCallTransactionParams & CommonTransactionParams): Transaction {
+}: ApplicationCallTransactionParams &
```
gmalouf, line 417: Noting this is a bit different than the other SDKs; if the `convertToAccess` equivalent is false, then one cannot pass holdings/locals at all - there isn't a separate `access` param.

src/transaction.ts
```diff
@@ -829,6 +980,13 @@ export class Transaction implements encoding.Encodable {
           this.applicationCall.appIndex
         )
       );
+      data.set(
+        'al',
+        resourceReferencesToEncodingData(
```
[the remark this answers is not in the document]
algorandskiy: correct, this is `toEncodingData` where fields are prepared (either in constructor or in deserialization).

src/makeTxn.ts
algorandskiy: please clarify, I probably did not get it - this params list has `access`:
```
  foreignApps,
  foreignAssets,
  boxes,
  convertToAccess,
  holdings,
  locals,
  access,
```

tests/5.Transaction.ts
```diff
@@ -2166,3 +2172,380 @@ describe('Sign', () => {
     });
   });
 });
+
+describe('Application Resources Refeneces', () => {
```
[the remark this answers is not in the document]
algorandskiy: fixed, thanks!

src/appAccess.ts
```diff
@@ -0,0 +1,288 @@
...
+    return accessList.length; // length is 1-based position of new element
+  }
+
+  const zeroAddr = Address.zeroAddress();
+  for (const rr of references) {
+    if (rr.address || rr.assetIndex || rr.appIndex) {
+      ensure(rr);
+    }
+
+    if (rr.holding) {
```
jannotti, line 49: Perhaps we should `continue` at the end of each `if (rr.xxxx) {` block? Or perhaps make these cascading `if else`s?

src/appAccess.ts
```diff
@@ -0,0 +1,288 @@
...
+    const address = item.get('d') as Address | undefined;
+    const assetIndex = item.get('s') as bigint | undefined;
+    const appIndex = item.get('p') as bigint | undefined;
+    if (address) {
+      references.push({ address });
+    } else if (assetIndex) {
+      references.push({ assetIndex });
+    } else if (appIndex) {
+      references.push({ appIndex });
+    }
```
jannotti, line 126: should we `continue` these pushes, so the reader knows there's nothing else going on?

src/appAccess.ts
```diff
@@ -0,0 +1,288 @@
...
+    const holding = item.get('h') as Map<string, unknown> | undefined;
+    if (holding) {
+      const hAddressIndex = ensureSafeUnsignedInteger(holding.get('d') ?? 0);
+      const hAssetIndex = ensureSafeUnsignedInteger(holding.get('s'));
+      if (!hAssetIndex) {
+        throw new Error(`Holding missing asset index: ${holding}`);
+      }
+      const hAddress =
+        hAddressIndex === 0
+          ? Address.zeroAddress()
```
jannotti, line 136: Does the zeroAddress become the sender somewhere else?

src/appAccess.ts
```diff
@@ -0,0 +1,288 @@
...
+      }
+    }
+    if (!addrFound) {
+      accessList.push({ address: addr });
+    }
+  }
+  function ensureAsset(asset: number | bigint) {
+    let assetFound = false;
+    for (const rr of accessList) {
+      if (rr.assetIndex && rr.assetIndex === asset) {
```
jannotti, line 235:
```suggestion
      if (rr.assetIndex === asset) {
```

src/appAccess.ts
```diff
@@ -0,0 +1,288 @@
...
+      }
+    }
+    if (!assetFound) {
+      accessList.push({ assetIndex: asset });
+    }
+  }
+  function ensureApp(app: number | bigint) {
+    let appFound = false;
+    for (const rr of accessList) {
+      if (rr.appIndex && rr.appIndex === app) {
```
jannotti, line 247:
```suggestion
      if (rr.appIndex === app) {
```

src/transaction.ts
```diff
@@ -180,6 +184,75 @@ function ensureBoxReference(input: unknown): TransactionBoxReference {
...
+    if (holding !== undefined) {
+      result = { ...result, holding: ensureHoldingReference(holding) };
+    }
+    if (locals !== undefined) {
+      result = { ...result, locals: ensureLocalsReference(locals) };
+    }
+    if (box !== undefined) {
+      result = { ...result, box: ensureBoxReference(box) };
+    }
+    return result;
```
jannotti, line 251: Since only one of these can be true, instead of accumulating into `result`, can you just return it?

src/appAccess.ts
algorandskiy: done

src/appAccess.ts
algorandskiy: it does not, it is kept as zero

src/appAccess.ts
algorandskiy: disagree, we do not want zero appIndex to appear in the list

src/appAccess.ts
algorandskiy: done

src/appAccess.ts
jannotti: This would only matter if `app` is 0, which means someone asked to convert a foreign array that contains a 0.  With this check in place `appFound` is going be false, and you'll end up pushing a `{appIndex: 0}` on line 263.

Your comment makes it sound like what you _don't_ want that though.

algorandskiy: right, I misread

jannotti (approved):

PR #1104, opened 2026-07-20
feat: remove dryrun due to removal in algod

joe-p:


v2_TO_v3_MIGRATION_GUIDE.md
```diff
@@ -316,9 +316,7 @@ The `IntDecoding.DEFAULT` option has been renamed to `IntDecoding.UNSAFE` in v3.
 
 ### Dryrun Utilities
 
-Dryrun-related utilities are deprecated because the `POST /v2/teal/dryrun` endpoint has been removed from algod. Prefer transaction simulation via `AlgodClient.simulateTransactions`.
-
-If you still need to parse legacy dryrun responses, use the fully-typed `modelsv2.DryrunResponse`/`modelsv2.DryrunTxnResult` models and the helper functions `dryrunTxnResultAppTrace` and `dryrunTxnResultLogicSigTrace`.
+Dryrun-related utilities have been removed because the `POST /v2/teal/dryrun` endpoint no longer exists on algod. This includes the `AlgodClient.dryrun` method, the `createDryrun`, `dryrunTxnResultAppTrace`, and `dryrunTxnResultLogicSigTrace` functions, and the `DryrunRequest`, `DryrunResponse`, `DryrunSource`, `DryrunState`, and `DryrunTxnResult` models. Use transaction simulation via `AlgodClient.simulateTransactions` instead.
```
iglosiggio, line 319: Is this the correct place to put the notice? (if it is it would be nice to add a "since v13" somewhere)

iglosiggio (approved):
LGTM

v2_TO_v3_MIGRATION_GUIDE.md
joe-p: I think it makes sense to put it here if someone is coming from v2 they'll want to know what happened to dry run. Since the API is tied to the go-algorand version and no AVM version, I'm going to omit the version for now but we can add it in if we want once it releases.

PR #1115, opened 2026-07-31
feat: support for consensus v42

joe-p:
This PR contains a few things

* Autogenerated types from the OAS bot PR
* Support for empty refs
* Add new state deltas
* Allow global schema and epp on UpdateApplication
* Add empty signers intended to be used with simulate


src/transaction.ts
```diff
@@ -247,6 +247,9 @@ function ensureResourceReference(input: unknown): TransactionResourceReference {
     if (box !== undefined) {
       return { box: ensureBoxReference(box) };
     }
+    // An empty resource reference is meaningful: it asks for a box I/O quota
+    // bump without naming a resource.
+    return {};
```
Argimirodelpozo: Should we check here that it is indeed an empty resource reference? otherwise something that is misconstrued/not one of the above and reaches here will be considered box I/O bump instead of throwing.

joe-p: Good catch. Fixed in 6adfd7b100290de7a8e248902f3eb52ca4bcc125

src/makeTxn.ts
```diff
@@ -539,7 +539,11 @@ export function makeApplicationCreateTxnFromObject({
 }
 
 /**
- * Make a transaction that changes an application's approval and clear programs
+ * Make a transaction that changes an application's approval and clear programs.
+ *
+ * The application's global state schema and extra program pages may also be changed during an
+ * update by providing numGlobalInts, numGlobalByteSlices, and/or extraPages. The local state schema
+ * cannot be changed after application creation.
```
Argimirodelpozo: Maybe we could add a quick comment here documenting the new sponsor mechanics as well

joe-p: Added in 285d004e852f517c6fea8a1b27da3f7522d4de2a

src/appAccess.ts
```diff
@@ -113,7 +113,12 @@ export function resourceReferencesToEncodingData(
           ],
         ])
       );
+      continue;
     }
+
+    // A reference that names no resource encodes as an empty ResourceRef, which
+    // the network reads as a request for a box I/O quota bump.
+    accessList.push(new Map<string, unknown>());
```
Argimirodelpozo, line 121: I think here we might have the same problem, although its harder to reach (but something like `assetIndex: 0` would end up here, as it wouldn't go into `ensure(.)` above)

Argimirodelpozo (dismissed):
LGTM!

src/appAccess.ts
joe-p: Hmmm yeah this method also doesn't seem to throw an error so introducing a throw may be unexpected. I think for now I'll leave it as is but we can keep an eye on it

Argimirodelpozo (approved):

PR #1124, opened 2026-08-10
CI: support beta releases

joe-p:
Enables beta releases of the SDK. It should be noted that I decided to do version parsing rather than adding a beta checkbox. It makes the workflow a bit more complex but protects us from accidentally releasing a beta version as the latest tag.


.github/workflows/create-release-pr.yml
```diff
@@ -104,31 +119,42 @@ jobs:
...
-          RELEASE_TAG: ${{ steps.set-release.outputs.release-tag }}
+          RELEASE_TAG: ${{ steps['set-release'].outputs['release-tag'] }}
+          IS_BETA: ${{ steps['set-release'].outputs['is-beta'] }}
         run: |
           rm -rf node_modules
           npm ci
           npm run build
           NEW_HASH=$(cat dist/browser/algosdk.min.js | openssl dgst -sha384 -binary | openssl base64 -A)
-          python3 scripts/bump_version.py ${RELEASE_TAG:1} --new_hash "${NEW_HASH}"
+          SKIP_README=()
```
onetechnical, line 129: This seems fine, but why make it an array at all?

onetechnical (approved):
Did not test it but it looks correct.

.github/workflows/create-release-pr.yml
joe-p: It's for the python side so it doesn't look like an empty string argument to argparse (source: claude)

